PR #9, opened 2016-02-09
Comment PR: Add comments for the design overview

arunoda:
Overall this is just great!

In my design, I was thinking to get invalidations to the client side (if needed) and decide to fetch or not).
But this is far better.

I add some more comments close to each section.
See `Files changed` tab

### DIFF
--- a/design/high-level-reactivity.md
+++ b/design/high-level-reactivity.md
@@ -59,7 +59,7 @@ Note that there is a GraphQL cache - what this does is decompose the results of 
 
 This is clearly very simplified, and you can read about how the Relay cache works on [Huey Petersen's website](http://hueypetersen.com/posts/2015/09/30/quick-look-at-the-relay-store/). The important thing to note is that the cache decomposes query results into a flat structure, and a smart cache will be able to generate a query to update just one list and its tasks, or just the list’s name, as necessary.
 
-// May be we can link Learn GraphQL here as well. Specially for someone who dislike GraphQL.
+// May be we can link Learn GraphQL. It's great, specially for a person who dislike GraphQL.
 // We've exprience in coverting them to lovers with LearnGraphQL :)
 // Edit: I'm waiting to add a section on Reactive GraphQL.
 
@@ -185,7 +185,7 @@ The rest of the document talks in terms of polling a dependency server for updat
 2. The application server could subscribe to the invalidations, and do the GraphQL queries _on the server_, and then diff that against the current state of the client and send a patch. This would make the system work almost exactly like Meteor today, and could be a good option for applications with fewer users and a great need for low latency.
 
 // I think it's better to go with Just the 1. We could optimize on top of that. It's not a good idea to maintain two different things.
-// If that app is small, it might be okay to change the polling interval smaller.  That'll do the trick.
+// If that app is small, it might be okay to change the polling interval smaller. That'll do the trick.
 
 Since both of those approaches don't change anything about the inherent design of the system and are relatively simple to implement, we'll leave them as optimizations for the future.
 
@@ -193,9 +193,10 @@ Since both of those approaches don't change anything about the inherent design o
 
 We haven’t talked at all about how the invalidation server finds out that a dependency version has incremented, and that clients should reload that data. At the lowest level, your code posts to the invalidation server the list of keys that need to be invalidated whenever you make a write to the data store. The next section will talk about a nice high-level wrapper around that for application developers.
 
-// Here, I'll be great we can submit the actual DB changes as well. I mean for simple quries. 
+// Here, I'll be great we can submit the actual DB changes as well. I mean for simple queries. 
 // Then we could do something with that. Specially in the client.
-// But then again, I get the issue of race-conditions when working with multiple servers.
+//
+// But then again, I get the issue of race-conditions, when working with multiple servers.
 // We clearly have no idea on which change actually applied to the DB. 
 // So, it's wise to forget this, at least for now :)
 
@@ -247,17 +248,18 @@ context.invalidateDependency(notificationDepKeyForUser(userId));
 
 Hopefully, with time, we can make more and more invalidations automatic, but it’s always good to have an “escape hatch” for more complex situations where the developer needs all the control they can get.
 
-// Add a section here: This is why we said, you need to add custom keys. Then we can have some examples like this:
+// Add a section here: This is why we said, you need to add custom keys. 
+// Then we can have some examples like this:
 
-1. Stop reactivity. Just publish a depKey which is random. We never will invalidate it.
-2. Counter.  Sometimes you don't need to it to be neal realtime reactive. So, we can throttle the invalidations and invalidate it for few minutes.
+1. Stop reactivity. Just publish a depKey which is random. Do not invalidate that key.
+2. Counter.  Sometimes you don't need to it to be neal realtime. So, we can throttle the invalidations and invalidate it for every few minutes.
 
 <img src="reactive-with-mutations-diagram.png" title="Reactive GraphQL with mutations" />
 
 You can see in the diagram how invalidations flow from the mutation to the relevant clients, which then refetch the data as needed.
 
 // How about outside invalidations. Lets say after a diret DB write from another non Meteor app.  
-// If we can connect to the invalidation server from outside that's awesome.
+// If we can connect to the invalidation server from outside that's awesome. (Again via a GraphQL mutation)
 
 ## Data drivers
 
@@ -294,8 +296,8 @@ The new system is designed to avoid this issue, and the implementation will be b
 
 After you analyze the two paths above, there should be a clear path to optimization through careful manual invalidations and disabling reactivity that will let you change a minimum of app code to "twist the knobs" on performance.
 
-> Oh Yeah! I'm seeing huge set of ideas. 
-> Just implement this. So, we can build a great APM and make money :P
+// Oh Yeah! I'm seeing huge set of ideas. 
+// Let's implement this. So, we can build a great APM and make money :P
 
 ## Implementation plan
 
--- a/design/high-level-reactivity.md
+++ b/design/high-level-reactivity.md
@@ -59,11 +59,17 @@ Note that there is a GraphQL cache - what this does is decompose the results of 
 
 This is clearly very simplified, and you can read about how the Relay cache works on [Huey Petersen's website](http://hueypetersen.com/posts/2015/09/30/quick-look-at-the-relay-store/). The important thing to note is that the cache decomposes query results into a flat structure, and a smart cache will be able to generate a query to update just one list and its tasks, or just the list’s name, as necessary.
 
+// May be we can link Learn GraphQL here as well. Specially for someone who dislike GraphQL.
+// We've exprience in coverting them to lovers with LearnGraphQL :)
+// Edit: I'm waiting to add a section on Reactive GraphQL.
+
 ## Reactive GraphQL
 
 People are excited about GraphQL, and it solves a lot of the problems which many developers, including Meteor’s customers and users, have brought to light. But when you’re building an app, you need more than just a way to query the server and get a response. For example, some parts of your app will want the data to update reactively when a different user makes a change.
 
 The client and server parts of the GraphQL system will need to collaborate to make this possible, but one of our goals is to make minimal changes to the current reference implementation of GraphQL so that developers can take advantage of current and future productivity tools for GraphQL.
+
+// Just Great.
 
 ### Dependencies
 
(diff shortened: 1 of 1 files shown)


design/high-level-reactivity.md
```diff
@@ -59,12 +59,18 @@ Note that there is a GraphQL cache - what this does is decompose the results of
 
 This is clearly very simplified, and you can read about how the Relay cache works on [Huey Petersen's website](http://hueypetersen.com/posts/2015/09/30/quick-look-at-the-relay-store/). The important thing to note is that the cache decomposes query results into a flat structure, and a smart cache will be able to generate a query to update just one list and its tasks, or just the list’s name, as necessary.
 
+// May be we can link Learn GraphQL. It's great, specially for a person who dislike GraphQL.
```
stubailo: Sure, if you want to link that in this doc for informational purposes I'd be happy to accept it. Just file a PR and I'll hit the button.

design/high-level-reactivity.md
```diff
@@ -59,12 +59,18 @@ Note that there is a GraphQL cache - what this does is decompose the results of
...
+// We've exprience in coverting them to lovers with LearnGraphQL :)
+// Edit: I'm waiting to add a section on Reactive GraphQL.
```
stubailo: Probably too early yet :P but we're going to build as fast as we can!

design/high-level-reactivity.md
```diff
@@ -94,6 +100,31 @@ Before, we gave an example of a GraphQL response for a simple query. If you make
...
+      _self: { key: '12341234', version: 3 },
+      lists: { key: '35232345', version: 4 }
+    }
+  }
+}
+```
+
+// Thinking about _self and lists. Is that because to avoid data re-fetching of lists when the username changes?
+// If so, what are the types we want to do it?
+// Is that something we can do it in the GraphQL Schema level.
```
stubailo: The `_deps` are specified at the schema level. The client also needs to request them, but that's easy because it has the schema from the introspection. Relay already does this to add the `id` field to your request even if you don't ask for it.

design/high-level-reactivity.md
```diff
@@ -94,6 +100,31 @@ Before, we gave an example of a GraphQL response for a simple query. If you make
...
+
+    __deps: {
+      _self: { key: '12341234', version: 3 },
+      lists: { key: '35232345', version: 4 }
+    }
+  }
+}
+```
+
+// Thinking about _self and lists. Is that because to avoid data re-fetching of lists when the username changes?
```
stubailo: Yes. So the idea is - if there is just the `_self` dep, it will refetch the whole subtree. If there is another one for a specific property, that will not be refetched. I think you have it right.

More details on this coming soon, we have a whole document about these deps, how they are returned, how they are refetched, etc. Hope to post that soon, maybe later this week, once I think about it some more, then we can discuss the details.

design/high-level-reactivity.md
```diff
@@ -94,6 +100,31 @@ Before, we gave an example of a GraphQL response for a simple query. If you make
 }
 ```
 
+// I think following example is simple and complete.
```
stubailo: Good idea. Can you file a PR to update the example?

design/high-level-reactivity.md
```diff
@@ -151,12 +184,22 @@ The rest of the document talks in terms of polling a dependency server for updat
 1. The invalidation server could accept websocket connections, and let the client subscribe to the dependency keys it cares about - this would mean the invalidation is pushed immediately, and then there is one roundtrip to fetch the actual data.
 2. The application server could subscribe to the invalidations, and do the GraphQL queries _on the server_, and then diff that against the current state of the client and send a patch. This would make the system work almost exactly like Meteor today, and could be a good option for applications with fewer users and a great need for low latency.
 
+// I think it's better to go with Just the 1. We could optimize on top of that. It's not a good idea to maintain two different things.
```
stubailo: I just want to emphasize that this design doesn't _stop_ you from implementing 2. I agree we shouldn't have it in the core implementation.

design/high-level-reactivity.md
arunoda: Great.

design/high-level-reactivity.md
```diff
@@ -151,12 +184,22 @@ The rest of the document talks in terms of polling a dependency server for updat
...
+// If that app is small, it might be okay to change the polling interval smaller. That'll do the trick.
+
 Since both of those approaches don't change anything about the inherent design of the system and are relatively simple to implement, we'll leave them as optimizations for the future.
 
 ### Invalidating dependencies
 
 We haven’t talked at all about how the invalidation server finds out that a dependency version has incremented, and that clients should reload that data. At the lowest level, your code posts to the invalidation server the list of keys that need to be invalidated whenever you make a write to the data store. The next section will talk about a nice high-level wrapper around that for application developers.
 
+// Here, I'll be great we can submit the actual DB changes as well. I mean for simple queries. 
+// Then we could do something with that. Specially in the client.
```
stubailo: Yes, you especially need to do that when invalidating arrays. We talked to the rocket chat people and this is basically what they do. You have to worry about race conditions, but if it's just appending to a list it's not a big deal. More details in the next doc coming out soon, we can have a good discussion around that.

design/high-level-reactivity.md
arunoda: :D

design/high-level-reactivity.md
arunoda: Sure.

design/high-level-reactivity.md
```diff
@@ -205,10 +248,19 @@ context.invalidateDependency(notificationDepKeyForUser(userId));
 
 Hopefully, with time, we can make more and more invalidations automatic, but it’s always good to have an “escape hatch” for more complex situations where the developer needs all the control they can get.
 
+// Add a section here: This is why we said, you need to add custom keys. 
+// Then we can have some examples like this:
+
+1. Stop reactivity. Just publish a depKey which is random. Do not invalidate that key.
```
stubailo: Yeah I was thinking in this case the dep key can just be a sentinel value, like `null` or something.

design/high-level-reactivity.md
arunoda: Yeah.

design/high-level-reactivity.md
```diff
@@ -205,10 +248,19 @@ context.invalidateDependency(notificationDepKeyForUser(userId));
...
+2. Counter.  Sometimes you don't need to it to be neal realtime. So, we can throttle the invalidations and invalidate it for every few minutes.
```
stubailo: Ah, you're saying have some parts of the subtree update slower than others. Neat idea!

design/high-level-reactivity.md
```diff
@@ -205,10 +248,19 @@ context.invalidateDependency(notificationDepKeyForUser(userId));
...
+
 <img src="reactive-with-mutations-diagram.png" title="Reactive GraphQL with mutations" />
 
 You can see in the diagram how invalidations flow from the mutation to the relevant clients, which then refetch the data as needed.
 
+// How about outside invalidations. Lets say after a diret DB write from another non Meteor app.  
+// If we can connect to the invalidation server from outside that's awesome. (Again via a GraphQL mutation)
```
stubailo: Of course! The invalidation component must have a clear API that anyone can call, so you can push changes from any microservice. This is crucial for all kinds of applications, like chat services. I'm excited to build a prototype to see how well this will work.

design/high-level-reactivity.md
arunoda: Got it.

design/high-level-reactivity.md
```diff
@@ -223,6 +275,16 @@ A first-class developer friendly backend data driver will need to:
 
 While the ideal driver would emit exact deps and invalidations automatically for all queries, this is not likely to be practical for an arbitrary datastore. In practice, drivers will likely fall back to broader deps and invalidations, and tooling will help developers identify places where these invalidations can be optimized.  The developer can then rework their query or emit manual invalidations as needed.
 
+// How about the design of the Invalidation Server. Here's my ideas:
+ * It would be great if it's stateless, so if we need more load we can just scale it.
```
stubailo: How will being stateless make it easier to scale? Do you mean if we don't have the stateful websocket connection? Or do you mean if we don't persist the version numbers?

design/high-level-reactivity.md
```diff
@@ -223,6 +275,16 @@ A first-class developer friendly backend data driver will need to:
...
+ * If not, we could shard it scale out. 
+ * This is not a thing we need to do now.
+ * But, we should plan for it.
+ 
+// Are we going to write it in Node.js
+
+I assume drivers need to do something in the invalidation later too. (Data fetching). So it's needed to do it in Node.js?
```
stubailo: The API between mutations and the invalidation server will be a websocket or HTTP, so you should be able to write it in whatever language you want at the end of the day.

design/high-level-reactivity.md
```diff
@@ -223,6 +275,16 @@ A first-class developer friendly backend data driver will need to:
 
 While the ideal driver would emit exact deps and invalidations automatically for all queries, this is not likely to be practical for an arbitrary datastore. In practice, drivers will likely fall back to broader deps and invalidations, and tooling will help developers identify places where these invalidations can be optimized.  The developer can then rework their query or emit manual invalidations as needed.
 
+// How about the design of the Invalidation Server. Here's my ideas:
+ * It would be great if it's stateless, so if we need more load we can just scale it.
+ * If not, we could shard it scale out. 
+ * This is not a thing we need to do now.
+ * But, we should plan for it.
+ 
+// Are we going to write it in Node.js
```
stubailo: The initial version will just be an NPM module you can use with Express, so you can run it in the same server, or on a separate server as a microservice. This seems like the simplest way to get started, and might also scale to reasonable production workloads. In the future when we hit scaling challenges we can rewrite it using whatever we want - Go, for example.

design/high-level-reactivity.md
arunoda: There is another person you should talk this. That's chatra. Guys. I will talk to them too.

design/high-level-reactivity.md
```diff
@@ -234,12 +296,20 @@ The new system is designed to avoid this issue, and the implementation will be b
 
 After you analyze the two paths above, there should be a clear path to optimization through careful manual invalidations and disabling reactivity that will let you change a minimum of app code to "twist the knobs" on performance.
 
+// Oh Yeah! I'm seeing huge set of ideas. 
+// Let's implement this. So, we can build a great APM and make money :P
```
stubailo: This is the money!

design/high-level-reactivity.md
```diff
@@ -234,12 +296,20 @@ The new system is designed to avoid this issue, and the implementation will be b
...
+
 ## Implementation plan
 
 Here's a diagram of all of the pieces we think will need to be built to have a complete system:
 
 ![Block diagram](block-diagram.png)
 
+// What does FlowRouter has to do with this. Do we need to do any changes?
```
stubailo: I just want to emphasize that we want to work with multiple routers. Right now Relay requires you to use Relay Router, we don't want to do that.

design/high-level-reactivity.md
```diff
@@ -234,12 +296,20 @@ The new system is designed to avoid this issue, and the implementation will be b
...
+
 It's a lot of stuff, but a lot of it already exists thanks to the Relay project, and some of the things can be contributed by the community once the structure is clearer, for example some of the database drivers.
 
+// Are we using Relay's connection API which is optimized for pagination. That's place where it adds the complexity to Relay and it's GraphQL part.
+// With our design, I don't think we need it.
```
stubailo: I don't see why we wouldn't just use the Relay thing if it works. Does it matter that it is complex under the hood if it's easy to use? Although I haven't looked at it in detail, it looks like Relay has a lot of really good stuff, and I'd rather use as much of it as possible.

Also, this will help us be friends with the Relay team.

design/high-level-reactivity.md
stubailo: Sounds great, would be awesome to make the connection.

design/high-level-reactivity.md
arunoda: If the user need it.

design/high-level-reactivity.md
arunoda: I mean how do we fetch data for the invalidations. 
Ah ha. Got it. You just need to refetch it from the client.
Invalidation server doesn't know about data.

stubailo: Exactly. This is why you will need to run a lot of application containers, and only one or very few invalidation server containers.

design/high-level-reactivity.md
arunoda: Part of this is a joke. We should build enough tools to get information out from the system.

stubailo: Yes, anyone should be able to get the information out easily. Then it's a race to build a great experience on top of the raw data.

design/high-level-reactivity.md
arunoda: It's not about its bad. It's good. 
But writing schemes with connection is kind hard.

That's something optional in Relay I think. Anyway, if you can up with a simple way to write schemas then there is no issue at all.

design/high-level-reactivity.md
arunoda: That's a good idea. Otherwise, this would lead to memory leaks in the invalidating server.

design/high-level-reactivity.md
arunoda: It's not WebSockets stateless. When it comes to scaling, if we can just scale invalidation severs and meteor servers as we need without worrying about anything else. 

That's what I meant.

design/high-level-reactivity.md
arunoda: Got it.

design/high-level-reactivity.md
arunoda: Okay. I mean the part that FlowRouter is a dark box. 

Hope we don't need to change anything on that.

stubailo: Yeah, for sure. We just need something you call in the `action` that lets you aggregate queries and stuff. It will need to work with SSR.

arunoda: Got it.

arunoda:
We are good here.

PR #55, opened 2016-06-21
Replacing slackin for Apollo

danahmadi:


danahmadi:
This replaces links to `http://slack.apollostack.com/` with our own overlay + slack invitation api calls. Also, we can link directly to the slack signup form via http://www.apollostack.com#slack

PR #128, opened 2018-02-21
Added Engine feature info & styling

johannakate:


johannakate:
I'm also attaching some other graphics I considered
persisted-queries
errors

PR #141, opened 2018-03-29
Add warning to documentation that the React tutorial is outdated

sryabkov:
A warning like the proposed one would have saved me three hours today. I kept trying to follow the tutorial and I kept running into errors thinking I was doing something wrong. But then it finally donned on me to look at the comments and I saw other people reporting the same problems. Having a warning here and ideally also at the linked blog post would help people who are trying to get up-to-speed on Apollo a lot of time and grief.

### DIFF
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -19,6 +19,7 @@ Getting started with Apollo is easy! You can follow these tutorials for step by 
 <h3 id="full-stack-graphql" title="Full stack tutorial"><a href="https://dev-blog.apollodata.com/full-stack-react-graphql-tutorial-582ac8d24e3b">Full Stack React and GraphQL Tutorial</a></h3>
 
 Jonas Helfer and friends take you through [building a simple GraphQL server and React app](https://dev-blog.apollodata.com/full-stack-react-graphql-tutorial-582ac8d24e3b) with Apollo.
+> The tutorial was written for Apollo 1.0 and has not yet been updated for Apollo 2.0
 
 <h3 id="howtographql">How to GraphQL</h3>
 


abernix:
Thank you for taking the time to raise this issue, @sryabkov, and apologies that this took you in the wrong direction!

cc @peggyrayzis @jbaxleyiii Is there a better direction we could steer developers at this stage of the docs: https://www.apollographql.com/docs/#getting-started?  Should we go with the disclaimer @sryabkov has suggested for now?

sryabkov:
@abernix 👍 Thanks for listening

@abernix @peggyrayzis @jbaxleyiii The next tutorial on the list (https://www.apollographql.com/docs/#howtographql) seems to be more up-to-date, at least judging by the apollo packages version numbers in package.json, so if you don't want to put any disclaimers on the page, at least consider moving it up, above https://www.apollographql.com/docs/#full-stack-graphql, because most people would probably start on the first tutorial listed, just like I did.

Ideally, I'd suggest that somebody do a quick review of documentation and flag all outdated (1.0) content in some kind of uniform fashion (which can also serve as a to-do list for documentation contributors)

peggyrayzis:
Thanks for reporting @sryabkov!

@abernix Instead of a disclaimer, I would rather point people to the getting started section in the React docs: https://www.apollographql.com/docs/react/essentials/get-started.html

What do you think?

abernix:
@sryabkov Would you have been okay if you had been taken to the [Getting Started guide](https://www.apollographql.com/docs/react/essentials/get-started.html) that Peggy mentioned above?  Were the problems related specifically to the `ApolloClient` constructor not being available, as many in the Medium comments had mentioned?  Do you feel like you would have _lost_ any information by not following Jonas' tutorial?

unicodeveloper:
I think this has been resolved.  Can I close this issue now @peggyrayzis?

abernix:
We've updated the Documentation home page to not include the link to this outdated tutorial and we've also updated the tutorial itself to disclaim that it's for an older version of Apollo Client.

Thanks for reporting this originally, @sryabkov! Much appreciated! :bowtie:

sryabkov:
👍

PR #155, opened 2018-05-29
Introduce guides, practices and fundamentals.

abernix:


docs/source/guides/schema-design.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Style conventions
+
+The GraphQL specification is flexible in the style that it dictates and doesn't impose specific naming guidelines. In order to facilitate development and continuity across GraphQL deployments, we suggest the following style conventions:
+
+* **Fields**: are recommended to be written in `camelCase`, since the majority of consumers will be client applications written in JavaScript.
+* **Types**: should be `PascalCase`.
+* **Enums**: should have their name in `PascalCase` and their values in `ALL_CAPS` to denote their special meaning.
+
+## Wrapping documents with gql
```
stubailo: This seems out of place in this article, since it's the only thing that has to do with the code. maybe we should have a separate article about schema implementation, with this and maybe schema organization

docs/source/guides/schema-design.md
```diff
@@ -0,0 +1,198 @@
...
+
+There are two common ways to write GraphQL schemas and queries. The first is to write queries into a `.graphql` file and import them into your other files for usage. The second is to write them wrapped them with the `gql` tag provided by the `graphql-tag` library.
```
stubailo: The most common one right now is by far to just have bare strings, so this paragraph seems inaccurate.

abernix: Agree this should be changed.  We shouldn't pretend like the most popular use case doesn't exist, but we have an increasing number of reasons to recommend using `gql`  to enable syntax-highlighting and to allow for future functionality.  It's now exported/available from `apollo-server` itself so this seems like an opportunity to mention that bare-strings are an option, but that we recommend using `gql` because of _x_, _y_ and _z_.

stubailo:
Note: the `/docs/` page is still the old one: https://deploy-preview-155--apollo-docs-index.netlify.com/docs/

Is that intentional?

docs/source/fundamentals/tips.md
```diff
@@ -0,0 +1,48 @@
+---
+title: Tips for success
+description: Your starting point for learning GraphQL and Apollo
+---
+
+TK WIP TODO finish this intro.  Depending on the size of an organization, there are usually existing practices surrounding the management of the data
```
peggyrayzis: For this intro, I think we want to instill confidence that we are the right people to guide teams on their GraphQL journey and that we've established these best practices based on years of experience from running GraphQL in production

docs/source/fundamentals/tips.md
```diff
@@ -0,0 +1,48 @@
...
+
+<h2 id="incremental-adoption">Incremental adoption</h2>
+
+While it makes an excellent longer-term goal, GraphQL doesn’t need to stand above the entire data tree in an organization in order to provide value and expending large amounts of engineering effort to switch an entire stack should generally be avoided.
+
+Unlike some technologies, which are difficult to embrace without getting buy-in from all the teams in an organization, GraphQL can be utilized by smaller units (e.g. component by component, team by team) and later integrated into the bigger picture.
+
+Thanks to GraphQL’s ability to fetch data from any data source, it makes sense to keep existing APIs as they are and use GraphQL to “wrap” the data.  This can be appreciated by organizations with existing collections of REST (or similar) APIs which power their user experiences since they can lean heavily on their existing investments.  This allows small teams to immediately realize many benefits of GraphQL without waiting for deeper adoption and allows existing versions of the application, which have already been deployed and rely on the existing API, to keep functioning uninterrupted.
+
```
peggyrayzis: I think we need a little more information on the how here. First, teams should start building their schema with one feature. They should develop the ideal API for that feature based on how they're consuming it on the client. Then, in their resolvers, they can call the resource associated with that feature + perform any data transformation that used to exist in Redux action creators. From there, they can start to integrate Apollo Client with the front-end code for that feature.

docs/source/fundamentals/tips.md
```diff
@@ -0,0 +1,48 @@
...
+<h2 id="schema">Schema scope and ownership</h2>
+
+APIs change as the products which use them evolve and their ideal structure is usually defined by the front-end applications which consume them.  While it’s certainly possible, and sometimes desirable, to have a single GraphQL schema which blankets an entire organization’s operational concerns, trying to manage a large schema presents challenges which are avoided by separating the schema into more manageable pieces that align with individual products.
+
+Luckily, the GraphQL Schema Definition Language (SDL) focuses on the shape of the data and not the actual implementation, allowing product teams to be the ideal custodians of their own GraphQL schemas.
+
+Having product teams own the GraphQL schema for their products allows the schema to act as an API contract within the team and enhances intra-team communication by providing clearly defined expectations.  These teams can utilize schema-driven development which lets them divide back-end and front-end work, mocking various parts of the API until full functionality is achieved, while remaining confident that they will arrive back at a unified solution.
```
peggyrayzis: It would be great if we could link to that article from the NYT CTO here where he admits that they should have started with the product when they first started transitioning to GraphQL. I think having a qualified voice to back up our claims would make this section even stronger.

docs/source/fundamentals/tips.md
```diff
@@ -0,0 +1,48 @@
...
+
+Organizations looking to offer a single API endpoint can assemble the individual product schemas, managed by individual product teams into a monolithic API by “stitching” the various schemas together.
+
+<h2 id="performance">Monitor GraphQL performance</h2>
+
+An API implemented using GraphQL allows developers to query the exact information they desire from an API and nothing more.  With proper visibility into how the API performs, developers can understand the implications of adding or removing fields, especially those which might be slow (notoriously or otherwise!).
+
+Apollo Engine is a GraphQL schema management and monitoring tool which provides insight on how a GraphQL API is behaving, and can proactively notify teams of degraded performance.
+
+When using Apollo Server with Apollo Engine, queries and mutations are enhanced with field-level tracing data which power views in Apollo Engine showing how removing a field from a query might speed up a particular component or, conversely, how adding a seemingly innocent field might actually do more harm than good!
```
peggyrayzis: A screenshot of tracing would be great here.

docs/source/fundamentals/benefits.md
```diff
@@ -0,0 +1,133 @@
...
+
+```bash
+GET /api/dogs/breeds
+GET /api/dogs/images
+GET /api/dogs/activities
+```
+
+With GraphQL, each query represents a single round trip from the client to server. If you'd like to reduce round trips even further, you can implement [query batching](https://www.apollographql.com/docs/engine/query-batching.html) to batch multiple queries into a single request. 
+
+<h2 id="dev-experience">Developer experience</h2>
```
abernix: I love this whole section `Developer experience` section and I feel like it should be higher in this portion of the guide!  Can we put it first?

docs/source/fundamentals/tips.md
```diff
@@ -0,0 +1,48 @@
...
+
+In addition to empowering front-end developers to make quick, educated decisions during development, the detailed metrics in Apollo Engine allow those who are responsible for data performance to quickly identify where improvements can be made and be made aware of bottlenecks or rising error rates if issues arise.
+
+This clarity allows pushing schema changes to production with the confidence of knowing that the API is performing as well as it was before, if not better!
+
+<h2 id="javascript">Write the server in JavaScript</h2>
+
+Facebook's reference implementation has been written in JavaScript since its original release and fresh developments in the GraphQL ecosystem have frequently appeared first in JavaScript or languages which transpile to JavaScript, like TypeScript.
+
+While existing back-end systems might already be implemented in other languages, a thin, JavaScript-based GraphQL implementation fits nicely in front of such systems and, due to the active maintenance of the JavaScript implementations, can bring the latest in GraphQL features to this existing infrastructure.
```
peggyrayzis: We should specifically mention Apollo Server here. Since we're the ones leading innovation in the GraphQL ecosystem with specifications like cache control and tracing, our server implementation is always the most up-to-date with the latest features. We could also mention schema stitching here, since many of the other language servers don't support it yet.

This section needs to answer why JavaScript (which you've done a great job at already) AND why Apollo Server (vs another JS server implementation).

docs/source/fundamentals/platform.md
```diff
@@ -0,0 +1,42 @@
+---
+title: The Apollo platform
+description: How Apollo helps you run GraphQL in production with confidence
+---
+
+The Apollo GraphQL data platform simplifies the data-layer within organizations, helps teams work more effortlessly with data from a variety of sources, and allows products to evolve faster.
```
peggyrayzis: This intro improved so much!! It's awesome 🎉

docs/source/fundamentals/platform.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Ready for production
+
+While the GraphQL specification was first made public by Facebook in 2015, GraphQL has been a key component of their mobile application deployment since 2012.
+
+At Apollo, we found GraphQL to be an excellent solution to many of the problems we encountered with existing techniques and now use it to power critical infrastructure.  Over the years, we’ve worked with the open-source community along with customers and partners of all sizes to continually bring new innovations to the open-source Apollo offerings and we're proud that those offerings are suitable for everything from startups to large-scale deployments.
+
+In addition to our own experience, we have received extensive feedback, contributions and support from enterprise customers who are actively using Apollo GraphQL in production.  A few of our most public and notable case-studies are:
+
+* **The New York Times**: Learn how The New York Times switched from Relay to Apollo & implemented features in their app such as SSR and persisted queries.
```
peggyrayzis: Let's add Airbnb to this list

peggyrayzis:
> Note: the /docs/ page is still the old one: https://deploy-preview-155--apollo-docs-index.netlify.com/docs/

@stubailo We're planning on redirecting the docs page to the first page of the fundamentals guide. We've moved all of the relevant existing content on that page to a new Platform section.

docs/source/fundamentals/benefits.md
```diff
@@ -0,0 +1,133 @@
...
+
+You’ll find that when you switch to Apollo Client, you’ll be able to delete a lot of unnecessary code related to data management. The exact amount will vary depending on your application, but some teams have reported up to thousands of lines. To learn more about how Apollo Client enables advanced features like optimistic UI, refetching, and pagination with less code, check out our guide on [state management](../guides/state-mgmt.html).
+
+<h2 id="performance">Improved performance</h2>
+
+In many cases, layering a GraphQL API over your existing REST endpoints can improve your app's performance, especially on devices with slow network connections. While you should always measure to determine how integrating GraphQL will affect your application, it's generally accepted that GraphQL improves performance by helping avoid round trips to the server and reducing payload size.
+
+<h3 id="payload">Smaller payloads</h3>
```
abernix: I understand and appreciate short(er) named anchors, but I don't think they should lose their clarity.  For example, from a maintenance standpoint if I had to look at `payload` in a list of, say, redirects, I'd rather it be called `smaller-payloads`, which is what this anchor would be if it the markdown renderer generated it from `##` syntax.  I feel that anyone using such anchor links elsewhere might also appreciate that, and it reduces the possibility of a conflict in the future (when we bring in, say, `## Compressed payloads` 🤷‍♂️)

docs/source/fundamentals/benefits.md
```diff
@@ -0,0 +1,133 @@
...
+
+<h3 id="data">Explore your API</h3>
+
+GraphQL's strongly typed query language allows developers to take advantage of incredible tooling for exploring GraphQL APIs. [GraphQL Playground](https://github.com/prismagraphql/graphql-playground) by Prisma is an excellent IDE featuring automatically generated docs for your schema and query execution with autocomplete. At a glance, you can see all the data available in your GraphQL API without diving into the backend code or knowing what source it came from.
+
+![GraphQL Playground](../assets/graphql-playground.png)
+
+Apollo Server 2.0 sets up GraphQL Playground out of the box, so you can start exploring your schema and making queries immediately. You also have access to GraphiQL within Apollo DevTools which is convenient for testing queries as you're working on front-end code with Apollo Client.
```
abernix: I feel like Apollo DevTools deserves its own `h4` here?  Perhaps GraphQL Playground could also become one, but the `h3` for "Explore your API" could start by explaining how all this is possible because of introspection?

docs/source/fundamentals/benefits.md
```diff
@@ -0,0 +1,133 @@
...
+const client = new ApolloClient({
+  uri: "https://dog-graphql-api.glitch.me/graphql"
+});
+```
+
+Teams who have switched to Apollo Client have reported [deleting thousands of lines of state management code](https://dev-blog.apollodata.com/reducing-our-redux-code-with-react-apollo-5091b9de9c2a) and lots of complexity from their application. Since Apollo Client supports managing both local and remote data, you can use the Apollo cache as a single source of truth for all global state in your application.
+
+<h3 id="tooling">Advanced, modern tooling</h3>
+
+Developing your GraphQL API with the Apollo platform unlocks advanced, modern tooling that helps teams uncover bugs quickly, gain visibility into their API, and develop advanced features such as caching with confidence.
```
abernix: Could we use this same sentence without the two occurrences of `advanced`?  I just don't like calling things `advanced` since it can be quite subjective.  I think this is just as powerful without it, and it leaves the level of appreciation up to the reader.

docs/source/fundamentals/benefits.md
```diff
@@ -0,0 +1,133 @@
...
+        name
+      }
+    }
+  }
+`;
+```
+
+Here, we're describing the shape of the object we want to receive from the server. GraphQL takes care of combining and filtering the data while returning exactly what we ask for.
+
+How do we use this query in our app? Apollo Client builds off of GraphQL's declarative approach to data fetching. In a React app, all of the logic for retrieving your data, tracking loading and error states, and updating your UI is encapsulated in a single Query component. This encapsulation makes composing your data fetching components with your presentational components a breeze! Let’s see how to fetch GraphQL data with Apollo Client in a React app:
```
abernix: Let's put "Query" inside backticks/graves to indicate it's a code reference (to the below snippet), à la `` `Query` `` and yielding `Query`.

docs/source/fundamentals/benefits.md
```diff
@@ -0,0 +1,133 @@
+---
+title: GraphQL benefits
+description: Why adopting GraphQL and Apollo will help you ship features faster
+---
+
+Managing data for modern applications is difficult. Developers have to aggregate data from multiple sources, distribute it upon multiple platforms, and plumb it into an app's UI. On top of that, front-end developers have to decide how to manage state on the client, all while executing complicated features such as caching and optimistic UI.
```
abernix: I think we can probably present the problem without calling it "difficult" (See my reasoning/similar concern with the use of "advanced" below).  For example: `Managing data in modern applications can present a number of challenges.`.

docs/source/fundamentals/platform.md
```diff
@@ -0,0 +1,42 @@
+---
+title: The Apollo platform
+description: How Apollo helps you run GraphQL in production with confidence
```
stubailo: I think this description should be made a bit more general, since Apollo as a whole isn't just about production - it also helps you get started

docs/source/fundamentals/platform.md
```diff
@@ -0,0 +1,42 @@
...
+
+The Apollo GraphQL data platform simplifies the data-layer within organizations, helps teams work more effortlessly with data from a variety of sources, and allows products to evolve faster.
+
+The more technical benefits of GraphQL adoption are well-documented in other areas of this documentation, however this “Fundamentals” section will focus on explaining pre-adoption recommendations which have been guided by Apollo’s many years of first-hand experiences working with everything from single-person apps to enterprise deployments.
+
+We’ll start by discussing the basics of GraphQL and Apollo and move into concepts, mindsets and tips which will help ensure a successful GraphQL story for those that follow it.
+
+## What is GraphQL?
+
+Fundamentally, GraphQL is a specification published by Facebook which defines concepts that enable a more structured data experience for products and teams alike.  It enables data from multiple sources to be unified into a single, strongly-typed API which can evolve over time and is easy for both humans and computers to explore.
```
stubailo: I wouldn't mention Facebook at all.

docs/source/guides/schema-design.md
```diff
@@ -0,0 +1,290 @@
...
+    "updateUserAge": {
+      "id": "1",
+      "age": "25",
+      "name": "jane doe"
+    }
+  }
+}
+```
+
+The first thing to note is that it’s most common to return the thing you’re updating from a mutation. In our example, we were updating a `User` record, so we returned that updated user. There’s nothing _enforcing_ this practice, but it’s highly recommended, because it’s often needed to have an updated user to let the clients update any local cache with a new instance of that `User`. For this same reason, it’s useful design mutations to update only one entity. If you wanted to update two unrelated entities, it’s recommended to in separate mutations. For more details on how to handle errors and warnings in mutations, see the [mutation responses](#mutation-responses) section below.
```
stubailo: > The first thing to note is that it’s most common to return the thing you’re updating from a mutation. In our example, we were updating a `User` record, so we returned that updated user.

I think a mutation should frequently return a special return type, like we detail below. And if the mutation does indeed represent an action that updates multiple types of things, it should probably return all of them. This sentence makes it seem like all mutations should just return a single object type, which I don't think is good advice.

> For this same reason, it’s useful design mutations to update only one entity. If you wanted to update two unrelated entities, it’s recommended to in separate mutations.

I actually feel the opposite of this - IMO a mutation should represent a user action on the frontend, regardless of how many entities that updates.

docs/source/guides/schema-design.md
```diff
@@ -0,0 +1,290 @@
...
+
+But what if we wanted to update more than just a single or couple attributes on a user? Passing each thing we need as a single argument would get tedious. Especially if multiple mutations used similar fields. For this, we can use input types, which are explained in the next section.
+
+<h3 id="mutation-input-types">Input types</h3>
+
+Input types are a special type in GraphQL which are defined as arguments to queries and, more commonly, mutations.  They can be thought of as object types for arguments, in addition to the other scalar types.  Input types are especially useful when multiple mutations require similar information; for example, when creating a user and updating a user require the same fields, like `age` and `name`.
```
stubailo: I think object types are actually most useful when a single mutation requires many arguments, and I'm not actually sure that reusing input types across mutations is a good idea. For example, in the `updateUser` case, you probably want most of the fields to be nullable, in case you want to update just some of them, but in the `createUser` case you might want to enforce that they are all present.

Even if that isn't the case at first, you might eventually need to change your arguments such that they do have different input requirements, and changing the input type of an argument is always a breaking change for existing clients (even if the fields are compatible).

So I'd suggest we change this recommendation to say:

1. We suggest input types for mutations which have a complex set of arguments
2. We suggest not sharing input types across multiple fields

docs/source/guides/schema-design.md
```diff
@@ -0,0 +1,290 @@
...
+  }
+}
+```
+
+Let’s break this down, field by field:
+
+* `code` is a string representing a transactional value explaining details about the status of the data change. Think of this like an HTTP status code.
+* `success` is a boolean indicating whether the update was successful or not. This allows a coarse check by the client to know if there were failures.
+* `message` is a string that is meant to be a human-readable description of the status of the transaction. It is intended to be used in the UI of the product.
+* `post` is added by the implementing type `AddPostMutationResponse` to return back the newly created post for the client to use!
```
stubailo: This doesn't match the example above, which has `user`

docs/source/guides/schema-design.md
```diff
@@ -0,0 +1,290 @@
...
+  name: String
+  age: Int
+  address: String
+  phone: String
+}
+```
+
+<h3 id="mutation-responses">Responses</h3>
+
+Mutations have a higher chance of causing errors than queries since they are modifying data.  A common way to handle errors during a mutation is to simply `throw` an error.  While that's fine, throwing an error in the resolver will return an error to the caller and prevent a partial response, which could be useful in the event of a partial update.  Consider the following mutation example, which tries to update a user's `name` and `age`:
```
stubailo: This is definitely a good reason to have a response object type and not just returning the object directly, but there are two more equally compelling reasons:

1. A mutation might actually end up changing multiple things, like for example favoriting a post might update that post _and_ the list of a user's favorite posts, meaning it should probably return both the post and the user
2. A mutation return value might acquire more fields over time, and if you just return the User directly, you won't have an opportunity to add any more fields there later.

I think it would be good to mention these up front, and say that only the simplest mutations should return a scalar or object type directly; probably most mutations should return a special object type tailored specifically for that mutation.

docs/source/guides/schema-design.md
```diff
@@ -0,0 +1,290 @@
...
+```graphql
+mutation updateUser {
+  updateUser(id: 1, user: { age: -1, name: "Foo Bar" }){
+    name
+    age
+  }
+}
+```
+
+With validation in place, this mutation might cause an error since the `age` is a negative value.  While it’s possible that the entire operation should be stopped, there’s an opportunity to partially update the user’s record with the new `name` and return the updated record with the `age` left untouched.
```
stubailo: My feeling is that it might be really confusing if we recommend people have mutations that can be partially successful. Is there any other reason to have mutations return this kind of status information?

docs/source/guides/schema-design.md
```diff
@@ -0,0 +1,290 @@
...
+Let’s break this down, field by field:
+
+* `code` is a string representing a transactional value explaining details about the status of the data change. Think of this like an HTTP status code.
+* `success` is a boolean indicating whether the update was successful or not. This allows a coarse check by the client to know if there were failures.
+* `message` is a string that is meant to be a human-readable description of the status of the transaction. It is intended to be used in the UI of the product.
+* `post` is added by the implementing type `AddPostMutationResponse` to return back the newly created post for the client to use!
+
+Following this pattern for mutations provides detailed information about the data that has changed and feedback on whether the operation was successful or not.  Armed with this information, developers can easily react to failures in the client and fetch the information they need to update their local cache.
+
+<h2 id="gql">Wrapping documents with `gql`</h2>
```
stubailo: I think this should be moved into a schema implementation section, which I heard Dhaivat might be writing.

docs/source/fundamentals/benefits.md
peggyrayzis: I'm cool with that 👍

docs/source/fundamentals/benefits.md
peggyrayzis: That's a really good point, I'll switch it to `smaller-payloads` for clarity. Also, I totally didn't realize that the markdown renderer autogenerates ids from `##` syntax! Thanks for the tip 😀

docs/source/fundamentals/benefits.md
peggyrayzis: Agreed! Just addressed all of these points and added a snazzy screenshot of the cache inspector for dev tools 👍

docs/source/fundamentals/tips.md
unicodeveloper: I think mentioning Apollo Server here makes sense. However, we can leave out Schema stitching for now.

docs/source/guides/schema-design.md
abernix: Addressed in 5c3d7b1ac4f068026a4c8c0976fb03b445013576.

docs/source/fundamentals/tips.md
hwillson: Does anyone have a link handy for that article? I can't seem to find it.

docs/source/fundamentals/tips.md
```diff
@@ -3,7 +3,9 @@ title: Tips for success
 description: Your starting point for learning GraphQL and Apollo
 ---
 
-TK WIP TODO finish this intro.  Depending on the size of an organization, there are usually existing practices surrounding the management of the data 
+Getting started with Apollo and GraphQL is quite easy. There are hundreds of tutorials available, put together by members of the amazing Apollo and GraphQL communities. While getting started is quite easy, there are some very important factors that should be considered upfront, to help deliver a successful Apollo GraphQL implementation. The Apollo team has been working with and on production GraphQL based projects for quite some time now, and has run into just about every gotcha and pitfall you can imagine happening. Over the past couple of years we've honed our GraphQL skills based on hard won experience, and a continuous desire to keep pushing this wonderful technology forward.
```
peggyrayzis: @hwillson this is looking good! Idk if I love the first two sentences because what if people don't think it's easy? I would maybe cut the first two and start with:

While you can get started with Apollo and GraphQL in less than ten minutes, how do you go from hello world to production? There are some very important decisions surrounding schema design, authorization, and state management that you'll have to make early on your journey to deliver a successful Apollo GraphQL implementation.

hwillson: Thanks @peggyrayzis - I went back and forth on that as well. Changing!

peggyrayzis:
LET'S DO THIS 🚀 :shipit:

PR #156, opened 2018-05-31
Address `schema-design.md` feedback.

abernix:
This PR targets #155 and hopefully addresses @stubailo's (excellent) [feedback](https://github.com/apollographql/apollo/pull/155#pullrequestreview-124634832).  Specifically, these:

* https://github.com/apollographql/apollo/pull/155#discussion_r191957939
* https://github.com/apollographql/apollo/pull/155#discussion_r191958434

Questionably addressed is:

* https://github.com/apollographql/apollo/pull/155#discussion_r191959082

And not addressed, yet:

* https://github.com/apollographql/apollo/pull/155#discussion_r191959210

### DIFF
--- a/docs/source/guides/schema-design.md
+++ b/docs/source/guides/schema-design.md
@@ -277,7 +277,7 @@ Let’s break this down, field by field:
 * `code` is a string representing a transactional value explaining details about the status of the data change. Think of this like an HTTP status code.
 * `success` is a boolean indicating whether the update was successful or not. This allows a coarse check by the client to know if there were failures.
 * `message` is a string that is meant to be a human-readable description of the status of the transaction. It is intended to be used in the UI of the product.
-* `post` is added by the implementing type `AddPostMutationResponse` to return back the newly created post for the client to use!
+* `user` is added by the implementing type `UpdateUserMutationResponse` to return back the newly created user for the client to use!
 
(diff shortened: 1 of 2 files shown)


docs/source/guides/schema-design.md
```diff
@@ -277,9 +258,86 @@ Let’s break this down, field by field:
...
+  title: String
+  "The textual body of the post."
+  body: String
+  "A list of URLs to render in the post."
+  mediaUrls: [String]
+}
+```
 
-Following this pattern for mutations provides detailed information about the data that has changed and feedback on whether the operation was successful or not.  Armed with this information, developers can easily react to failures in the client and fetch the information they need to update their local cache.
+Input types can also be used when different operations require the exact same information, though we urge caution on over-using this technique since changes to `input` types are breaking changes for all operations which utilize them.
```
peggyrayzis: This looks great! Maybe we could add a sentence here cautioning readers about reusing input types for queries and mutations since it's likely that some fields would be nullable on one but not the other?

abernix:
Followed up with https://github.com/apollographql/apollo/pull/156/commits/7a09ad19de31ca10858cb18c642de08afdf4cbcb, https://github.com/apollographql/apollo/pull/156/commits/b56da257a8d3ed41942df67d67c8a7e2675bac5f and https://github.com/apollographql/apollo/pull/156/commits/8e2e2b78835398ba56582105d5077db07b6d0370.  Thank you for the feedback!

PR #161, opened 2018-06-01
Schema Organization Article

Poincare:
Covers schema and resolver modularization, type extensions, context object and a brief primer on models.


peggyrayzis:
Closing, this page would be more appropriate in the server docs.

PR #164, opened 2018-06-05
React Apollo Testing

JakeDawkins:
# Testing React Apollo 

This is a guide specifically for testing React Apollo. Based off of @excitement-engineer's (great) original work [in this PR](https://github.com/apollographql/apollo-client/pull/3309).

**A few notes:**
- I'm keeping this separate from server testing, since combined I think they'll be too long. **I'm open to discussion on this, though**.
- I added the guide here rather than React Apollo's docs, since I feel like we have a good set of `guides` here, and would hate for this to be missing. I also feel like the view logic may transfer to other UI integrations, but I don't have a lot of insight into that.
- I went with using `wait(0)` instead of refactoring and counting renders like some other exmples do, for simplicity. I wanted to lower the barrier of entry for testing as much as possible.
- I used `react-test-renderer` instead of Enzyme or anything else, just to keep the toolset as simple as possible.

### DIFF
--- a/docs/source/guides/testing-react-apollo.md
+++ b/docs/source/guides/testing-react-apollo.md
@@ -1,0 +1,305 @@
+---
+title: Testing React Apollo
+description: Have peace of mind when using React-Apollo in production
+---
+
+TODO: Intoduction
+
+Running tests against code meant for production has long been a best practice. It provides additional security for the code that's already written, as well as prevents accidental regressions in the future. Components utilizing React Apollo are no exception.
+
+Although React Apollo has a lot going on under the hood, the library provides multiple tools for testing that simplify those abstractions, and allows complete focus on the component logic. These testing utilities have long been used to test the React Apollo library itself, so developers can trust in their stability and long-term support.
+
+## An introduction
+
+React Apollo relies on [context](https://reactjs.org/docs/context.html) in order to pass the Apollo Client instance through the React component tree. In addition, React Apollo makes network requests in order to fetch data. This behavior affects how you write tests for components that use React Apollo.
+
+This guide will explain step-by-step how you can test your React Apollo code. The following examples will be using the [Jest](https://facebook.github.io/jest/docs/en/tutorial-react.html) testing framework, but most concepts should be reusable with other libraries. These examples also show the basic [Test renderer](https://reactjs.org/docs/test-renderer.html) from the React library.
+
+Consider the example below:
+
+```js
+import React from 'react';
+import gql from 'graphql-tag';
+import { Query } from 'react-apollo';
+
+// make sure the query is also exported -- not just the component
+export const GET_DOG_QUERY = gql`
+  query getDog($name: String) {
+    dog(name: $name) {
+      id
+      name
+      breed
+    }
+  }
+`;
+
+export const Dog = ({ name }) => (
+  <Query query={GET_DOG_QUERY} variables={{ name }}>
+    {({ loading, error, data }) => {
+      if (loading) return 'Loading...';
+      if (error) return `Error! ${error.message}`;
+
+      return (
+        <div id="dog-info">
+          {data.dog.name} has breed {data.dog.breed}
+        </div>
+      );
+    }}
+  </Query>
+);
+```
+
+If we were to try and write a test for the `<Dog />` component then you will get an error that the `client` is missing in the context.
+
+```js
+// Broken because it's missing Apollo Client in the context
+it('should render without error', () => {
+  renderer.create(<Dog />);
+});
+```
+
+In order to fix this we could wrap the component in an `ApolloProvider` and pass an instance of Apollo Client to the `client` prop. However, this will cause our tests to run against an actual backend which makes the tests very unpredictable for the following reasons:
+
+- The server could be down
+- There may be no network connection
+- The results are not guaranteed to be the same for every query
+
+```js
+// Not predictable
+it('renders without error', () => {
+  renderer.create(
+    <ApolloProvider client={client}>
+      <Dog name="Buck" />
+    </ApolloProvider>,
+  );
+});
+```
+
+## MockedProvider
+
+To test the component in true isolation, mock all calls to the GraphQL endpoint. This makes the UI consistent every time tests are run.
+
+React Apollo provides the `MockedProvider` component in the `react-apollo/test-utils` library to do just that! `MockedProvider` allows specifying the exact results that should be returned for a certain query using the `mocks` prop.
+
+Here's an example of a test for the above `Dog` component using `MockedProvider`:
+
+```js
+// dog.test.js
+
+import { MockedProvider } from 'react-apollo/test-utils';
+
+// The component AND the query need to be exported
+import { GET_DOG_QUERY, Dog } from './dog';
+
+const mocks = [
+  {
+    request: {
+      query: GET_DOG_QUERY,
+      variables: {
+        name: 'Buck',
+      },
+    },
+    result: {
+      data: {
+        dog: { id: '1', name: 'Buck', breed: 'bulldog' },
+      },
+    },
+  },
+];
+
+it('renders without error', () => {
+  renderer.create(
+    <MockedProvider mocks={mocks}>
+      <Dog name="Buck" />
+    </MockedProvider>,
+  );
+});
+```
+
+This file shows how to define the mocked response for a query. The `mocks` array takes objects with specific `request`s and their associated `result`s. In this example, when the provider receives a `GET_DOG_QUERY` it returns the object under the `result` key.
+
+<!-- PICK UP HERE WITH TALKING ABOUT LOADING STATE -->
+
+Despite having mocked the request the above test will still fail. React Apollo will first show a loading state before it successfully retrieves the data and renders the dog name.
+
+In order to overcome this issue you can separate the React Apollo components from presentational components and test these independently.
+
+Let's see how we can refactor the code to improve testability.
+
+```js
+export const GET_DOG = gql`
+query getDog($name: String) {
+  dog(name: $name) {
+    id
+    name
+    breed
+  }
+}
+`;
+
+// The React Apollo component
+export const DogQuery = ({ children, name }) => (
+  <Query query={GET_DOG} variables={{name}}>
+    {({ loading, error, data }) =>
+      children({
+        loading,
+        error,
+        dog: data && data.dog
+      })
+    }
+  </Query>
+);
+
+// The presentational component
+export const DogInterface = props => {
+  const { loading, error, dog } = props;
+  if (loading) return "Loading...";
+  if (error) return `Error! ${error.message}`;
+
+  return <div>{dog.name} has breed {dog.breed}</div>;
+};
+
+// Connecting the components together.
+export const Dog = props => (
+  <DogQuery name={props.name}>{result => <DogInterface {...result} />}</DogInterface>
+);
+```
+
+Let's go back to our test file and see how we can test the `<DogQuery />` component.
+
+```js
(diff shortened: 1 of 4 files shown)


docs/source/guides/testing-react-apollo.md
```diff
@@ -0,0 +1,332 @@
...
+it('renders without error', () => {
+  renderer.create(
+    <MockedProvider mocks={mocks} addTypename={false}>
+      <Dog name="Buck" />
+    </MockedProvider>,
+  );
+});
+```
+
+This example shows how to define the mocked response for a query. The `mocks` array takes objects with specific `request`s and their associated `result`s. In this example, when the provider receives a `GET_DOG_QUERY` witht the specified variables, it returns the object under the `result` key.
```
hwillson: Typo (extra "t"): witht

docs/source/guides/testing-react-apollo.md
```diff
@@ -0,0 +1,332 @@
...
+
+### addTypename
+
+You may notice the prop being passed to the MockedProvider called `addTypename`. The reason this is here is because of how Apollo Client normally works. When a request is made with Apollo Client normally, it adds a `__typename` field to every object type requested. This is to make sure that Apollo Client's cache knows how to normalize and store the response. When we're making our mocks, though, we're importing the raw queries _without typenames_ from the component files.
```
hwillson: Maybe backtick `MockedProvider` here to be consistent.

docs/source/guides/testing-react-apollo.md
```diff
@@ -0,0 +1,332 @@
...
+    },
+  };
+
+  const component = renderer.create(
+    <MockedProvider mocks={[dogMock]} addTypename={false}>
+      <Dog name="Buck" />
+    </MockedProvider>,
+  );
+
+  await wait(0); // wait for response
```
hwillson: Should we mention where `wait` is coming from? Maybe a quick mention that it's coming from https://www.npmjs.com/package/waait in a code comment, or show `import wait from 'waait';` in the code sample.

hwillson:
Just to add - I saw your comment about ignoring it for now, but I was too excited! 🙂

docs/source/guides/testing-react-apollo.md
JakeDawkins: added an import for this example, and mentioned it in the following paragraph. Great idea 👍

docs/source/guides/testing-react-apollo.md
```diff
@@ -0,0 +1,334 @@
...
+
+This example shows how to define the mocked response for a query. The `mocks` array takes objects with specific `request`s and their associated `result`s. In this example, when the provider receives a `GET_DOG_QUERY` with the specified variables, it returns the object under the `result` key.
+
+### addTypename
+
+You may notice the prop being passed to the `MockedProvider` called `addTypename`. The reason this is here is because of how Apollo Client normally works. When a request is made with Apollo Client normally, it adds a `__typename` field to every object type requested. This is to make sure that Apollo Client's cache knows how to normalize and store the response. When we're making our mocks, though, we're importing the raw queries _without typenames_ from the component files.
+
+If we don't disable the adding of typenames to queries, our imported query won't match the query actually being run by the component during our tests.
+
+> In short, If you don't have `__typename`s in your queries, pass `addTypename={false}` to your `MockedResolver`s.
```
excitement-engineer: I am wondering what the `addTypename` prop is actually used for. When testing I simply want to link some data to the imported query without having to deal with typenames like so:

```js
import { Query } from “./some_component”;

const mocks = [{
    request: {
         query: Query
    },
    result: someData
}]

renderer.create(
<MockedProvider mocks={mocks}>
   <Component />
</MockedProvider>
);
```

I actually created a PR in the past where I simplified the `MockedProvider` (see [#1882](https://github.com/apollographql/react-apollo/pull/1882)) and removed the `addTypename` prop. Back then I created this PR because I was constantly running into the issue that the MockedProvider was not returning data for my query (this was because `addTypename` was true by default). 

Furthermore, I have a [PR]](https://github.com/apollographql/react-apollo/pull/1995) open that sets the `addTypename` prop to false by default because this will allow developers to simply create mock data for their queries without having to worry about typenames.

Is there a specific use case that the `addTypename` prop enables? I have personally never had to use it in my projects and it has actually gotten in my way when testing apollo components. So I am wondering if more users will run into this.

Given that the MockedProvider has not been documented yet, I think that this may be a good moment to flesh out this API before it becomes public knowledge. Thoughts?

docs/source/guides/testing-react-apollo.md
```diff
@@ -0,0 +1,334 @@
...
+    },
+  };
+
+  const component = renderer.create(
+    <MockedProvider mocks={[dogMock]} addTypename={false}>
+      <Dog name="Buck" />
+    </MockedProvider>,
+  );
+
+  await wait(0); // wait for response
```
excitement-engineer: Does this always work consistently? As I remember correctly adding a timeout of 0 doesn't always guarantee that the loading state is gone. Therefore in the `react-apollo` [examples](https://github.com/apollographql/react-apollo/pull/1994/files) I used [wait-for-expect](https://github.com/TheBrainFamily/wait-for-expect) to test the components and this proved to work great:) What do you think?

```js
import wait from "wait-for-expect";

it("tests", () => {
// This statement will wait until loading UI disappears before continuing
await wait(() => {
    isNotLoading();
})

// Data is present 

});
```

docs/source/guides/testing-react-apollo.md
JakeDawkins: I think I agree that `false` is a more sensible default.

I think we left it to `true` by default to make sure that it wouldn't be a breaking change to anyone using it currently. We hadn't documented it, but I know there are some people that we had pointed to this code initially, and may be using it in their projects.

Thoughts @jbaxleyiii?

docs/source/guides/testing-react-apollo.md
JakeDawkins: Oh, interesting. I'll have to check out those examples and see. I never saw the loading state not be gone (although these examples are very simple). 

Otherwise, I like wait-for-expect. I'll check it out and see if it's a simple enough solution!

docs/source/guides/testing-react-apollo.md
excitement-engineer: Oh ok, if people are actively using it then a breaking change may not be such a good idea. With the future release of react-apollo v3 then perhaps this behavior can be changed. Thanks for clarifying:)

@jbaxleyiii  Do you know by any chance why the `addTypename` prop was actually added in the first place? I'm wondering what use cases it enables

docs/source/guides/testing-react-apollo.md
excitement-engineer: Great, thanks a lot for taking it into consideration:)

docs/source/guides/testing-react-apollo.md
JakeDawkins: Yeah, v3 may be a good place to reconsider this. 

Originally, the mocked provider acted like the normal provider did, and added typename by default. We added this prop to disable that

docs/source/guides/testing-react-apollo.md
```diff
@@ -0,0 +1,334 @@
...
+  button.props.onClick(); // fires the mutation
+
+  await wait(0);
+
+  const tree = component.toJSON();
+  expect(tree.children).toContain('Deleted!');
+});
+```
+
+Testing UI components isn't a simple issue, but hopefully with these tools, you can feel more confident when testing components dependent on data.
```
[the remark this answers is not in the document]
JakeDawkins: Ideally, yeah! I wrote a test for doing that in the codesandbox example I linked to at the bottom.

I commented it out because for some reason when testing, it would throw the error on the preview window (even though the component is never mounted).

I didn't have the time to figure out what was going on there, unfortunately

JakeDawkins:
@excitement-engineer @hwillson I've moved this PR to https://github.com/apollographql/apollo-client/pull/3572 and added notes for the things still left to fix/discuss.

Also, sorry @excitement-engineer was right in opening the PR there originally 🙈

abernix:
I realize there has been some punting back and forth going on here, and it's partially my fault for not acting on this a bit sooner (also, not at all @JakeDawkins' fault), but after following some other conversations, and with the future of the more-general (as contrasted with the separate server and client docs) `/docs` site (this repository's content) on my mind, I'm going to re-open this PR and close https://github.com/apollographql/apollo-client/pull/3572 — not because of any mistake that @excitement-engineer has made, but due to a lack of clarity (on our part) on where it should live.

That's not to say that the back-and-forth isn't understandable though since we haven't historically had a home for content which "spans the stack" and this PR's content (testing) has two faces: a client and a server side.

While it might make sense to co-locate the respective sections within the (separate) server and client sections of the documentation, this separation creates three different entry-points for subjective content which would be more easily found within the (new) Guides section we now have on [`/docs`](https://www.apollographql.com/docs/).  Also, this pattern of putting client testing best-practices within "Recipes" (as was proposed with https://github.com/apollographql/apollo-client/pull/3572), is a little less reusable since some guides will have independent client and server sections which do not overlap (e.g. "Testing", which currently has two non-contiguous sections) and others will have end-to-end conversation and implementation details (e.g. _Access Control_).

If we don't put this opinionated content into a single place, we'll also make it difficult (for ourselves and others) to decide where new content should land since a guide might start as client-only and graduate into a more full-stack recommendation.  This would necessitate the relocation of this content somewhat organically and unexpectedly, which would be less than ideal.

We've already planted the seeds for having implementation guides outside of documentation, and by putting this testing content in the new "Guides" section of `/docs`, we further support this investment and drive developers to a single resource for opinionated best-practices.  Lest we forget that the [Apollo Server documentation](https://www.apollographql.com/docs/apollo-server/v2/) doesn't have a corresponding "Recipes" section for the upcoming "Server testing" guide!

Of course, we can (and should!) provide links to this content from the server and client guides, but having the content live there doesn't seem manageable long-term.  As we get more and more content which has similar client/server division that "Testing" has, perhaps we can accommodate it better with some navigational element (Server/Client tabs? 🤷‍♂️ ), but for now, I think we should be comfortable having a single "Testing" guide with "Client" and "Server" sections, or if the content is prohibitively long, separate "Testing: Client" and "Testing: Server" guides.

I think I've outlined some very compelling reasons above, but of course, I'm happy to discuss this further.  Ultimately, my goal here is to optimize and streamline the documentation experience.  I can't wait to get this excellent content into a new, highly-visible home, and I'm convinced that the Guide section on `/docs` is the place for that.

unicodeveloper:
Totally agree with you @abernix 👍

unicodeveloper:
@JakeDawkins @hwillson @peggyrayzis @abernix Are we considering any other change or addition to this content. If not, then we can merge it in ASAP! And then continue iteration on it as new recommendations or examples or updates come in.

JakeDawkins:
Added a note about potentially needing to wait longer than `wait(0)` for some components, and another about mutation errors. There's an example in the codesandbox, so I think this should be fine!

docs/source/guides/testing-react-apollo.md
```diff
@@ -0,0 +1,340 @@
+---
+title: Testing React Apollo
```
stubailo: We don't currently use the words "React Apollo" as a brand name in that order. Maybe "testing react with apollo" or "testing your frontend"?

docs/source/guides/testing-react-apollo.md
```diff
@@ -0,0 +1,340 @@
+---
+title: Testing React with Apollo
```
stubailo: This makes it sound like Apollo is a testing tool.

unicodeveloper:
Please merge @abernix

PR #169, opened 2018-06-13
Editorial for https://github.com/apollographql/apollo/pull/164.

abernix:
This is editorial work for https://github.com/apollographql/apollo/pull/164.  It also targets that PR and will merge into it.

The changes I've introduced here try to remove the subject of "you".  This is a pattern which we've tried to perpetuate in other portions of the documentation.

In my experience this has generally made the content more succinct (i.e. shorter), but also simplifies translation in some languages and makes the content more easily digestible for those who might be less proficient by removing the extra, unnecessary subject.

I have some concerns about using the [`awwait`](https://npm.im/awwait), since it's basically a full npm module that does:

    await (new Promise(r => setTimeout(r, 0)))

I think we can probably accomplish the same testing without this module by using `Promise`-aware test tooling which waits for fulfillment of the `Promise`s before testing.  We can iterate on this later though, and this is a great start.

Oh, I also made a number of changes of the term "resolves" to "is fulfilled" since the term "resolves" would indicate that the execution of the `Promise` didn't chain didn't result in an error - which is entirely possible.

### DIFF
--- a/docs/source/guides/testing-react-components.md
+++ b/docs/source/guides/testing-react-components.md
@@ -5,11 +5,11 @@ description: Have peace of mind when using react-apollo in production
 
 Running tests against code meant for production has long been a best practice. It provides additional security for the code that's already written, and prevents accidental regressions in the future. Components utilizing `react-apollo`, the React implementation of Apollo Client, are no exception.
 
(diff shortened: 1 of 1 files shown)


JakeDawkins:
I'm definitely in favor of replacing `waait` with that one-liner or restructuring tests later to not need it at all.

I also didn't think about the translation issues or understanding issues of using 2'nd person in writing 👍

PR #178, opened 2018-07-01
Add Performance & File Uploads Guide.

unicodeveloper:
This PR adds a Performance Guide which includes Automatic Persisted Queries and CDN Integration.

This is better suited here than the Apollo Server Docs.


JakeDawkins:
Do you think we should have an example that shows a full implementation, rather than just commenting out resolver instructions?

unicodeveloper:
I guess you are referring to the file uploads right? A full implementation will be good as an example repo or glitch to link to. Just as we linked to the examples at the bottom of the page.

PR #182, opened 2018-07-09
Fix link

avin:
Fixed link to `State management` chapter


evans:
@avin Thanks so much!

PR #191, opened 2018-07-20
Update non-breaking change use case

goldcaddy77:
The way it is written, there are 2 changes in the example, which could confuse users.  Instead, we keep the signature of the first param the same.

- [x] docs


JakeDawkins:
Hey @goldcaddy77! Thanks for the PR. I'm going to close it, because this was intended. 

In the example, the purpose of the field was changed to be able to look up users by a list of ids _or_ by group _or_ both, so the Ids arguments is no longer a required field. That part wouldn't be a breaking change, either, since it wouldn't affect clients already using the field.

PR #192, opened 2018-08-03
Change `authentication` to `authorization`

souserge:
In the first example, the author used `req.headers.authorization` to store/retrieve the authentication token, but in the second code snippet, `req.headers.authentication` was used

### DIFF
--- a/docs/source/guides/access-control.md
+++ b/docs/source/guides/access-control.md
@@ -62,7 +62,7 @@ To do this kind of authorization, we can just modify the context function.
 ```js
 context: ({ req }) => {
  // get the user token from the headers
- const token = req.headers.authentication || '';
+ const token = req.headers.authorization || '';
 
  // try to retrieve a user with the token
  const user = getUser(token);


souserge:
I would actually change the name to `authentication` in both cases, since it is what the token is supposed to provide. What do you think?

fbartho:
I'm concerned that this example is problematic anyhow.
1. Shouldn't `headers` be an instance of [`class Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) -- 
2. Shouldn't we encourage people to use [`Headers.get()`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/get)
3. Shouldn't we encourage people to use the canonical casing for that header `Authorization` instead of relying on some library to have called `toLowerCase()`?

souserge:
Judging by the rest of their documentation, the code snippets aren't intended to be copy-pasteable. But I agree, @fbartho, it would be great for them to be consistent with the real world.

JakeDawkins:
Thanks for the PR @souserge! That must've been overlooked. 

As for those points, @fbartho, those may have been intentional, because of the support of the serverless integrations. I'm not 100% sure.

PR #202, opened 2018-08-21
Update graphql-glossary.md

diana-lakatos:
- [ ] feature
- [ ] blocking
- [x] docs

Suggesting an edit for clarity. Although now there is some word repetition, it's much clearer that built-in and custom directives are different, and what "it" referred to. I think repetition in this case improves comprehension.


JakeDawkins:
I love it! Thanks for making this clearer @diana-lakatos!

PR #205, opened 2018-08-27
Explain how to mock GraphQL errors

YoruNoHikage:
I was testing my application and had to look in a specific part of documentation to understand that `graphQLErrors` are just taken from the result. This might help a few people like in https://github.com/apollographql/react-apollo/issues/1127.

Using `error` key on the mock will only result in a `networkError` define. This fix explains how to define GraphQL errors which might not be explicit when a beginner is not used to it.


JakeDawkins:
Agreed @YoruNoHikage! Thanks for clarifying 👍

PR #212, opened 2018-10-03
Add content for data sources section

unicodeveloper:

### DIFF
--- a/docs/source/tutorial/data-source.md
+++ b/docs/source/tutorial/data-source.md
@@ -381,3 +381,5 @@ In the `getUsersByLaunch` method, a `launchId` is accepted via the method argume
 So, if the `userIds` variable is empty, then the `getUsersByLaunch` method returns an empty array. However, if the `userIds` variable is not empty, then users with those ids are fetched from the users table.
 
 If no users were found, the `getUsersByLaunch` method returns an empty array, else an array of users with their respective `id`, `email` and `avatar` is returned!
+
+In the next section of this tutorial, we'll write our resolvers!
--- a/docs/source/tutorial/data-source.md
+++ b/docs/source/tutorial/data-source.md
@@ -41,13 +41,15 @@ class MvrpAPI extends RESTDataSource {
 module.exports = MvrpAPI;
 ```
 
-The `https://mvrp.herokuapp.com/api/` endpoint is a simple REST API that returns data for cars. Furthermore, the `MvrpAPI` class implementation in the code above contains a `getAllCars` and `getACar` functions that wrap convenience methods provided by the `RESTDataSource` class for performing HTTP requests. In this example, the built-in `get` method used is responsible for `GET` requests.
+The `https://mvrp.herokuapp.com/api/` endpoint is a simple REST API that returns data for cars. Furthermore, the `MvrpAPI` class implementation in the code above contains a `getAllCars` and `getACar` functions that wrap convenience methods provided by the `RESTDataSource` class for performing HTTP requests.
+
+In this example, the built-in `get` method used is responsible for `GET` requests. The `RESTDataSource` class provides other methods like `post`, `delete`, and `patch` for `POST`, `DELETE`, and `PATCH` requests respectively.
 
 Now that you have an understanding of how data sources work, let's hook it up for our tutorial app.
 
 Create a new `datasources` folder inside the `src` directory. This folder will contain our data source files. Now, create `launch.js` within the `datasources` directory.
 
-The REST API endpoint we'll use for our app is `https://api.spacexdata.com/v2/`. Add the endpoint as a base URL as shown in the code below:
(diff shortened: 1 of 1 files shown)


docs/source/tutorial/data-source.md
```diff
@@ -3,6 +3,424 @@ title: "2. Hook up your data sources"
...
+
+In the code above, we required the `RESTDataSource` class and extended it with our custom `LaunchAPI` class. We then set up the base URL in the class constructor to the endpoint that fetches the data needed for our application.
+
+The next step is to add methods to the `LaunchAPI` class that corresponds to the type of queries our UI will fetch from the server. According to our app specifications, we'll need to get all the launches, and get a specific launch. So, let's take care of the former immediately.
+
+_src/datasources/launch.js_
+
+```js
+...
+async getAllLaunches() {
```
peggyrayzis: Can you please update this example to the latest version? I think we're using a `launchReducer` function instead of mapping it directly.

docs/source/tutorial/data-source.md
```diff
@@ -3,6 +3,424 @@ title: "2. Hook up your data sources"
...
+      },
+      launchSuccess: launch.launch_success,
+    };
+  }) : [];
+}
+```
+
+In the code above, `this.get('launches')`, makes a `GET` request to `https://api.spacexdata.com/v2/launches` and stores the returned data in the `res` variable. If the `res` variable is not empty, then the `getAllLaunches` method returns an object that corresponds with the schema fields of the `Launch` schema type, else, an empty array is returned.
+
+Let's refactor the `getAllLaunches` method to be a lot cleaner and concise. Copy the `launchReducer` method below and add to the file. Now, refactor the `getAllLaunches` method to use the `launchReducer` method as shown below:
```
peggyrayzis: I see where you're going with this, but I think we should teach them the `launchReducer` first instead of dragging out the example/

docs/source/tutorial/data-source.md
```diff
@@ -3,6 +3,424 @@ title: "2. Hook up your data sources"
...
+    launchIds.map(launchId => this.getLaunchById({ launchId })),
+  );
+}
+```
+
+The `getLaunchById` method takes in a flight number and returns the data for a particular launch, while `getLaunchesByIds` returns several launches based on their respective `launchIds`. `Promise.all()` takes an array of promises and returns a single promise that resolves when all the promises in the array have been resolved with their fulfilled values.
+
+<h2 id="database">Connect a database</h2>
+
+A data store is needed for saving and fetching user information. It's also important for user trips. Let's make use of [SQLite](https://www.sqlite.org) for our app's database. SQLite is a self-contained, light-weight, zero-configuration and embedded SQL database engine.
```
peggyrayzis: We're going to take care of all the Sequelize code for them since we don't want the tutorial to focus too much on this part. Please slim this section down so it just focuses on the Apollo specific parts.

docs/source/tutorial/data-source.md
```diff
@@ -3,6 +3,424 @@ title: "2. Hook up your data sources"
...
+}
+```
+
+In the `isBookedOnLaunch` method, we invoke the `findAll` method of the `trips` table to find out if the particular launch passed to the method via `launchId` has been booked by the logged-in user.
+
+In the various methods that we created and copied to the `UserAPI` class, you must have noticed `this.store.users` and `this.store.trips`. These are two tables from our potential SQLite data store. Let's create the store!
+
+<h4 id="create-the-store">Create the Store</h4>
+
+Create an `utils.js` file inside the `src` directory. Now, copy the code below and add to it.
```
peggyrayzis: Remove all of the code related to creating a store -- we're going to have them fork a copy of the repo that already has this code filled in for them :)

docs/source/tutorial/data-source.md
```diff
@@ -3,6 +3,424 @@ title: "2. Hook up your data sources"
...
+// Set up Apollo Server
+const server = new ApolloServer({
+  typeDefs,
+  dataSources: () => ({
+    launchAPI: new LaunchAPI(),
+    userAPI: new UserAPI({ store }),
+  }),
+  context: async ({ req }) => {
+
+    const email = 'johndoe@apollo.com';
```
peggyrayzis: Remove hard coded email

docs/source/tutorial/data-source.md
unicodeveloper: This section was done intentionally @peggyrayzis. The next code sample after this showed them why I did it this way and introduced them to a simple refactoring approach that birthed the `launchReducer` function.

docs/source/tutorial/data-source.md
unicodeveloper: I have replaced this with a filler link to the supposed code the readers should fork.

docs/source/tutorial/data-source.md
unicodeveloper: Done

PR #214, opened 2018-10-11
Run your graph in production

unicodeveloper:
This is the first draft for the `Run your graph in production` section.

The content here covers:

* Deploying your graph API to Now
* Publishing your schema to Engine via the Apollo CLI
* How to hook up Apollo Server with Engine for performance monitoring

The content here doesn't cover:
* Operation Registry
* VS code set up

### DIFF
--- a/docs/source/tutorial/production.md
+++ b/docs/source/tutorial/production.md
@@ -93,7 +93,9 @@ Set the `ENGINE_API_KEY` environment variable via the command line:
 $ ENGINE_API_KEY=YOUR_API_KEY npm start
 ```
 
-Once this is done, you can now have access to all the features that Apollo Engine provides:
+Once this is done, run a query in GraphQL Playground, and log in to your Engine account to monitor the graph's performance.
+
+Apollo Engine provides all of the following features highlighted below:
 
 * **Schema Explorer:** With Engine's powerful schema registry, you can quickly explore all the types and fields in your schema with usage statistics on each field. This metric makes you understand the cost of a field. How expensive is a field? Is a certain field in so much demand?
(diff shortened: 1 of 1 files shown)


docs/source/tutorial/production.md
```diff
@@ -3,8 +3,104 @@ title: "4. Run your graph in production"
...
+
+Now looks for a `start` script in the `package.json` file to start the app. Once the script is present, Now will be able to start the app.
+
+Go ahead and run the `now` command from the root directory of the app.
+
+```bash
+$ now
+```
+
+The `now` command immediately deploys your graph API to the cloud and returns the hosted URL. The returned URL comes in this format: `<NOW_APP_NAME>.now.sh`.
```
peggyrayzis: Isn't there a configuration setting to specify the app name? If so, let's use that.

docs/source/tutorial/production.md
```diff
@@ -3,8 +3,104 @@ title: "4. Run your graph in production"
...
+
+Once you publish your schema to Apollo Engine, it becomes the basis for comparison for validating future schemas and avoiding breaking changes. Therefore, a schema should be re-published to Apollo Engine each time a new schema is deployed.
+
+Go ahead and install the `apollo` CLI via npm:
+
+```bash
+npm install --global apollo
+```
+
+To publish a schema, start your GraphQL server and run the command below:
```
peggyrayzis: Before they publish, they need an Engine API key. Can you please add instructions on how to get that key?

unicodeveloper: I added instructions on how to get the API key on line 53

docs/source/tutorial/production.md
unicodeveloper: There isn't. It's a random generated unique name because we are using a free plan and the readers too would be using that.

docs/source/tutorial/production.md
unicodeveloper: Moved the instructions up a bit!

PR #228, opened 2018-11-07
Add a mention about apollo-link-http

YoruNoHikage:
I wasn't sure about the need to remove the link http since upload client is also a terminating link.
After struggling a bit, I stumbled on https://stackoverflow.com/a/49536534.

Also, I find it better to collocate the apollo boost warning with the install instruction, since it's a pre-requisite to go further.


peggyrayzis:
Thanks for your contribution! We've removed file uploads from the platform docs, so I'm closing this PR.

PR #241, opened 2018-11-13
Initial Client Awareness Documentation

evans:
This lays out the initial client awareness documentation with some pictures

https://deploy-preview-241--apollo-docs-index.netlify.com/docs/platform/client-awareness.html


docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
-description: How to configure Apollo Engine for client awareness
----
\ No newline at end of file
+description: What is client awareness and how to add it to the Apollo Platform
+---
+
+Client identity is central to the Apollo Platform and enables tracking how all
+the consumers use the data graph. The Apollo Platform allows segmenting usage
+data by client name and version. Filtering by client information provides a
+field-level understanding of how the users interact with the GraphQL api in
```
unicodeveloper: api => API

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+real-time. In addition to per client metrics, understanding this granular
+detail informs how the GraphQL schema can evolve and reacts to a new client
+release.
+
+## Use Cases
+
+Often a GraphQL api is used by multiple consumers with different frequencies,
```
unicodeveloper: api => API

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+version, tracking these requests across the stack and enabling filtering on
+both dimensions. This segmentation provides an understanding of which queries
+and fields are required by clients. Conversely, this breakdown also indicates
+which clients are important based on relative usage.
+
+![client overview](../img/client-awareness/overview.png)
+
+In addition to understanding how clients use a GraphQL api, the Apollo Platform
+enables effective isolation and rapid triage for issues that affect a single
+client or portion of clients begin to experience issues, by filtering out of
```
unicodeveloper: Something is off with this paragraph. 

> ...that affect a single client or portion of clients begin to experience issues,..

Please rephrase

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+the functional clients. In the opposite sense, if a certain client becomes
+problematic, such as requesting expensive fields or using deprecated fields,
+the Apollo Platform enables tacking down the misbehaving client to start a
```
unicodeveloper: tacking => tracking

unicodeveloper: Maybe we can use another word in place of "misbehaving".   "Faulty" perhaps?

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+conversation with the owner and move toward solving the issue. When changing,
+replacing, or deprecating a field in the api, the client metadata enables
+quickly identifying down the client side changes that need to occur.
```
unicodeveloper: > ..enables quickly identifying down the client side changes that need to occur.

"down" should be removed. It's not needed here.

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+
+## Setup
+
+Client awareness is a full stack solution that threads client information from
+the consumer to server, so we need to add some code to the server and client.
+
+### Server
+
+The server is responsible for collecting and assigning the client information
+to a request.. To provide metrics to the Apollo Platform, pass a
```
unicodeveloper: > to a request..

Please remove the extra full stop

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+
+server.listen().then(({ url }) => {
+  console.log(`🚀  Server ready at ${url}`);
+});
+```
+
+### Client
+
+The client or consumer of the GraphQL api is responsible for including the
+information in a way that the server understands.In this case, we add the
```
unicodeveloper: There should be space between `understands.` and `In this case,`

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
 ---
 title: Identifying clients
-description: How to configure Apollo Engine for client awareness
----
\ No newline at end of file
+description: What is client awareness and how to add it to the Apollo Platform
+---
+
+Client identity is central to the Apollo Platform and enables tracking how all
+the consumers use the data graph. The Apollo Platform allows segmenting usage
```
zionts: The text here is great! When looking at the Netlify preview, it was a little overwhelming of a wall of text, though, and I think that **bolding** specific pieces would be useful. This sentence in particular is a good one to bold.

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+data by client name and version. Filtering by client information provides a
```
zionts: s/Filtering by client information/Filtering by client

docs/source/platform/client-awareness.md
zionts: I'm not sure I understand why this bit about field-level understanding. I also think that we should keep our terminology consistent and stick to either `client` or `consumer` rather than introducing another concept of "users"

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+field-level understanding of how the users interact with the GraphQL api in
+real-time. In addition to per client metrics, understanding this granular
```
zionts: s/per client/per-client

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+detail informs how the GraphQL schema can evolve and reacts to a new client
```
zionts: parallelism: should be "informs how the GraphQL schema can evolve and react to new client releases"

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+release.
+
+## Use Cases
+
+Often a GraphQL api is used by multiple consumers with different frequencies,
+subselections, and permissions. The Apollo Platform accepts client name and
+version, tracking these requests across the stack and enabling filtering on
```
zionts: This sentence sounds a little strange to me. Maybe something like "The Apollo Platform allows specification of client name and version with all reported metrics to enable filtering those reported metrics to a specific client or set of clients"
don't need to add, but across different stacks as well :O holy moly!

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+both dimensions. This segmentation provides an understanding of which queries
+and fields are required by clients. Conversely, this breakdown also indicates
```
zionts: nit: s/required/used

zionts: Also I'd remove the "Conversely"

Maybe enumeration is a useful tool for engagement in docs? So you can say "this segmentation provides two things: 1. ____, 2. _____" 🤷‍♂️

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+which clients are important based on relative usage.
+
+![client overview](../img/client-awareness/overview.png)
```
zionts: this image is great! Maybe we could even put it at the top of the file?

docs/source/platform/client-awareness.md
zionts: Agreed -- this sentence doesn't sound quite right. What are you trying to say here?

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+
+Client identity is central to the Apollo Platform and enables tracking how all
+the consumers use the data graph. The Apollo Platform allows segmenting usage
+data by client name and version. Filtering by client information provides a
+field-level understanding of how the users interact with the GraphQL api in
+real-time. In addition to per client metrics, understanding this granular
+detail informs how the GraphQL schema can evolve and reacts to a new client
+release.
+
+## Use Cases
```
zionts: Because this section is about particular use cases, I really think that enumeration would be helpful. I also think that it might be better for the use cases to be below the setup, but open to other opinions. We should motivate the usage of course, but I think that could be accomplished with an impactful photo.

docs/source/platform/client-awareness.md
zionts: I really like this whole motivating paragraph, and I think it belongs in an "Overview" section rather than a "Use cases" section

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+
+![client field](../img/client-awareness/field-usage.png)
+
+Similarly to deprecation, additions to a GraphQL api often means that a client will change. These client changes can be done incrementally or discretely during a cutover period. The cutover period and time immediately following change the utilization of the GraphQL api drastically and can expose some unexpected behavior. Filtering by client version enables monitoring the health of a release as it occurs. The following demonstrates a cutover from one backend to another.
+
+![druid cutover](../img/client-awareness/cutover.png)
+
+## Setup
+
+Client awareness is a full stack solution that threads client information from
```
zionts: end clause sounds a little informal for my taste. Maybe "so configuration is required in both client and server code"

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+the consumer to server, so we need to add some code to the server and client.
+
+### Server
+
+The server is responsible for collecting and assigning the client information
+to a request.. To provide metrics to the Apollo Platform, pass a
+`generateClientInfo` function into the `ApolloServer` constructor. The
+following checks the GraphQL query `extensions` for a `clientInfo` field and
```
zionts: It's probably worth a `>NOTE:` or something for the query `extensions` since that isn't actually a part of the GraphQL spec and our readers may be confused.

zionts: I think this is a good section for /fullTraces implementor leads. Can we say something like "If you're using Apollo Server 2.0 or higher, you don't need to modify any server code to pass along client information. The default behavior is _____. To override the default behavior, specify _______.

If you're NOT using Apollo Server and would like to gain client awareness, please [reach out](mailto: adam@apollographql.com) to work with us to add support to your server language of choice.

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+Similarly to deprecation, additions to a GraphQL api often means that a client will change. These client changes can be done incrementally or discretely during a cutover period. The cutover period and time immediately following change the utilization of the GraphQL api drastically and can expose some unexpected behavior. Filtering by client version enables monitoring the health of a release as it occurs. The following demonstrates a cutover from one backend to another.
+
+![druid cutover](../img/client-awareness/cutover.png)
+
+## Setup
+
+Client awareness is a full stack solution that threads client information from
+the consumer to server, so we need to add some code to the server and client.
+
+### Server
```
zionts: This should definitely specify the default.

docs/source/platform/client-awareness.md
```diff
@@ -1,4 +1,117 @@
...
+  console.log(`🚀  Server ready at ${url}`);
+});
+```
+
+### Client
+
+The client or consumer of the GraphQL api is responsible for including the
+information in a way that the server understands.In this case, we add the
+client name and version to the `extensions` field. We do this by defining a
+custom `ApolloLink`:
```
zionts: Isn't there also just a way to do this in the newest Apollo Client release in the constructor? If not, can we add one? It should also be specified that this is being matched up with the server code. Maybe we should have the server setup piece at the bottom for an "advanced" use case since most people probably don't need to know about the overridability?

zionts:
I'm going to hold off on giving this another pass until we add the constructor argument to Apollo Client tomorrow 😍 

cc @hwillson

PR #242, opened 2018-11-13
Set up redirects for guides to new platform links

unicodeveloper:

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -51,3 +51,13 @@ url: https://www.apollographql.com/docs
 root: /docs/
 
 public_dir: public/docs
+
+redirects:
+  /docs/guides/schema-design.html:
+    docs/platform/schema-design
+  /docs/guides/security.html:
+    docs/platform/operation-registry
+  /docs/guides/monitoring.html:
+    docs/platform/integrations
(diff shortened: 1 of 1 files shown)


docs/_config.yml
```diff
@@ -51,3 +51,13 @@ url: https://www.apollographql.com/docs
...
 
 public_dir: public/docs
+
+redirects:
+  /docs/guides/schema-design.html:
+    docs/platform/schema-design
+  /docs/guides/security.html:
+    docs/platform/operation-registry
+  /docs/guides/monitoring.html:
+    docs/platform/integrations
```
abernix: The `redirects` in this `_config.yml` are all client-side redirects (handled in [this code](https://github.com/meteor/meteor-theme-hexo/blob/376001e3b68ecbcaa49f2c98cd8c75929324862b/layout/layout.ejs#L203-L228)), but none of those pages that are being redirected _from_ exist anymore.   It would still be possible to use client-side redirects if those pages existed, but it's still better for search engines if we did this on the server.

Therefore, all of the above redirects should go into https://github.com/apollographql/apollo/blob/master/docs/public/_redirects and follow the format of that file `/source /dest`.

PR #231, opened 2018-11-08
Included the "isEmail" library import

samie820:
It was not stated that the `isEmail` library should be imported before it can be used.

PR #252, opened 2018-11-19
Update resolvers.md

fzhukov:
add required module in Auth chapter

without this in my case i take error "Context creation failed: isEmail is not defined"


daniman:
Closing because I merged #231

PR #254, opened 2018-11-19
Fix button styles

trevorblades:
This branch fixes some button styles on mobile devices and reorganizes some of the CSS for those elements.

I also adjusted the media queries around these buttons so that they don't break to their smaller version until the screen gets much narrower (was 990px, now it's 480px.)

**Before**
screen shot 2018-11-19 at 11 43 52 am

**After**
screen shot 2018-11-19 at 11 44 22 am


peggyrayzis:
LGTM 👍 thanks for making this fix!

PR #258, opened 2018-11-21
Update client awareness docs to use latest AC

evans:
We update the setup docs to use the latest Apollo Client constructor


evans:
Currently this needs to wait on Apollo-server@2.2.3 for https://github.com/apollographql/apollo-server/pull/1960

PR #259, opened 2018-11-22
Fix no-service-found issue when `apollo service:check`

yetsun:
When running `npx apollo service:check`
https://www.apollographql.com/docs/tutorial/production.html#publish,
It will throw the no-service-found error as follows:
```
  ✔ Loading Apollo Project
  ✖ Checking service for changes
    → No service found to link to Engine
Error: No service found to link to Engine
    at Task.task (~/.nvm/versions/node/v8.0.0/lib/node_modules/apollo/lib/commands/service/check.js:16:31)
    at Promise.resolve.then.then.skipped (~/.nvm/versions/node/v8.0.0/lib/node_modules/apollo/node_modules/listr/lib/task.js:167:30)
    at <anonymous>
```
The reason is`apollo.config.js` is missed to config the service name.

The side effect of this change is that it forces the user to name their
service as `space-explorer`, which is later used in the client,
so I think it's better to enforce the service name to avoid confusing 
and unnecessary error.


docs/source/tutorial/production.md
```diff
@@ -15,16 +15,26 @@ In a production application, you should set up this publishing script as part of
 
 <h3 id="api-key">Get an Engine API key</h3>
 
-First, we need an Apollo Engine API key. Navigate to [Apollo Engine](https://engine.apollographql.com/), login, and click on New Service at the top. The prompt will instruct you to name your service. When you're finished, click Create Service. You'll see a key appear prefixed by `service:`. Copy that key so we can save it as an environment variable.
+First, we need an Apollo Engine API key. Navigate to [Apollo Engine](https://engine.apollographql.com/), login, and click on New Service at the top. The prompt will instruct you to name your service. Name the new service `space-explorer`. When you're finished, click Create Service. You'll see a key appear prefixed by `service:space-explorer:`. Copy that key so we can save it as an environment variable.
```
abernix: We'll need to figure out another pattern here since the name must be globally unique.  That's to say, we can't recommend `space-explorer` here since once it's used (and it probably already is, by now) the next user will not be able to use it.

JakeDawkins:
I don't think this change is strictly necessary. The service name can be parsed from the api key, so the service name in the apollo config shouldn't be required (although we do have a config file in the final project).

PR #260, opened 2018-11-22
Change Buffor constructors to `from` method

wojciech-bilicki:
This pull request updates the way we use constructor, to get rid of the deprecation warning.


evans:
@wojciech-bilicki Thank you so much!

PR #261, opened 2018-11-23
Update to walkthrough: Deploy schema to Engine

bwobst:
When deploying a new schema the following fails: `apollo service:check`. It is able to load the apollo project, but fails at checking the service for changes. It returns this error: `Error: No service found to link to Engine`

I needed to use the snippet from Engine to deploy it:
```
npx apollo schema:publish --endpoint=http://localhost:4000/graphql --key="service:my-api-key"
```

Also, when I tried to use the original `npx apollo service:check && npx apollo service:push` command in a new terminal I had to set the engine api key envar in that window before running the command.


evans:
With the added clarification from #291, I think we can close this. @bwobst thank you so much for the PR!

PR #262, opened 2018-11-23
Update client.md

bwobst:
Client does not start up without cache or link properties defined.

### DIFF
--- a/docs/source/tutorial/client.md
+++ b/docs/source/tutorial/client.md
@@ -58,13 +58,19 @@ _src/index.js_
 
 ```js
 import { ApolloClient } from "apollo-client";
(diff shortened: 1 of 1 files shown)


sw-yx:
a lot of the docs have been written with Apollo Client 3.0 in mind, where apollo-boost has been merged into core. however it hasn't yet been done, not even in alpha, so this is the problem when actually trying out whats in the docs. to fix, just import and use from apollo-boost for now.

```js
import ApolloClient from 'apollo-boost';

const client = new ApolloClient({
  uri: 'http://localhost:4000/graphql' // no error
});
```

evans:
@bwobst Thank you so much!

anztrax:
@sw-yx  thanks 👍

PR #263, opened 2018-11-25
Add docs for the metrics reporting endpoint

### LINKED ISSUE #277: Add docs for the metrics reporting endpoint
Enable other server implementors to integrate with Apollo Engine's reporting functionality. This should essentially describe the format for reporting metrics as well as other resources in creating a metrics reporting agent like Apollo Engine Reporting.

zionts:
Fixes #277


zionts:
I think this is most of the way there, but it would benefit from some images and perhaps better organization. Either way, ready for eyes and thoughts :)

PR #265, opened 2018-11-26
Update _redirects

FredyC:
I am not sure if it's exactly the document that should be there, but the specified one does not exist anymore.


daniman:
Thanks @FredyC!

PR #272, opened 2018-12-01
update schema tutorial

cdnil:
Because the argument name is talked above (missionPatch argument), I just remove the argument names for legibility.


evans:
@cdnil Thank you!

PR #273, opened 2018-12-01
Fix typo

agrimm:
Incorrect markdown left an asterisk on the page.


agrimm:
The CLA is asking for my mailing address. Never mind.

PR #279, opened 2018-12-06
fix link typo on mutation subheading

hagnerd:
The link on the `Mutation` subheading was pointing back up to the `Object & Scalar Types` heading.


daniman:
Thanks @hagnerd!

PR #281, opened 2018-12-08
Update apollo command instructions

trevor-scheer:
This PR updates some documentation around using the apollo command. For a new project, `service:check` will fail, so we should only recommend that after we know they've published once.

Also adds in the endpoint flag since there's currently a PR open to remove the default, which this relies on: https://github.com/apollographql/apollo-tooling/pull/809


trevor-scheer:
Looks like my auto-format hit the MD doc, happy to revert that change, though it looks harmless.

PR #285, opened 2018-12-14
Update the Apollo Config docs

daniman:
This PR moves the Apollo config docs to the `References` section from the `Platform` section.

It reorganizes the content in the doc to delineate what's required and what's optional in the Apollo config, making it more like an API reference doc.

I thought this doc could be the source of truth for config options and that we could sprinkle tidbits of when you'd need to add/edit your config in our other docs, but ultimately be able to point users back here as a reference.


docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+    },
+  },
+};
+```
+
+### `includes`, `excludes`
+
+*Optional* –– by default, Apollo tools will look under a `./src` directory to find all operations and SDL to extract.
+
+Client projects often contain client-side schema definitions for local state with Apollo Client. To make sure the Apollo CLI and VS Code extension can find these files and read them correctly, you may need to add some additional definitions in your config file. To configure which folders to include, and which to ignore, adjust your config like so:
```
trevor-scheer: We might also state that includes and excludes are both arrays of glob patterns.

Two thoughts:

1. I'm tempted to mark this as required so people don't accidentally skim this. Missing `includes` might be the biggest gotcha I've seen up to this point, and for now the error messaging around it isn't great.

2. Maybe it's fine so long as the messaging is improved (when no files are found).

I think #2 seems reasonable, but I'd like for us to mark it as required _until_ the messaging is improved. Could you let me know what you decide on? If this, then I'll open an issue against tooling to add better messaging and come and update these docs to say optional.

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+
+```js line=3-4
+module.exports = {
+  client: {
+    includes: ['./imports/**/*.js'],
+    excludes: ['**/__tests__/**/*'],
```
trevor-scheer: We should note the `excludes` default somewhere as well:
`excludes: ["**/node_modules", "**/__tests__"]`

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+module.exports = {
+  client: {
+    tagName: 'graphql',
+  },
+};
+```
+
+### `addTypename`
+
+*Optional* –– Apollo will by default add the `__typename` field to all operations automatically.
```
trevor-scheer: We should recommend that this option be 1:1 with how their `ApolloClient` is configured. We might also say directly that this will add `__typename` to all the generated types during codegen, since right now that's all this field is used for.

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+  client: {
+    clientOnlyDirectives: ['connection', 'type'],
+    clientSchemaDirectives: ['client', 'rest'],
+  },
+};
+```
+
+`clientOnlyDirectives` are directives that should be stripped out of the operation before being sent to the server. An example of this is the `@connection` directive.
+
+`clientSchemaDirectives` are directives that indicate a portion of the operation that is not meant to be sent to the server. These directives are removed as well as the fields they are placed on. An example of this type of of directive is the `@client` directive.
```
trevor-scheer: >of of

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+
```
trevor-scheer: We should mention these fields on the `client` key as well:
```js
  name?: ClientID;
  referenceID?: string;
  version?: string;
```
For a command like `client:push`, these will provide the fine-grained identification in Engine 😄

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+```js line=2
+module.exports = {
+  service: { ... },
+};
+```
+
+Setting up a `service` in your Apollo config will save you from needing to use flags your Apollo CLI commands like `apollo service:push` and `apollo service:check`. You can set up the schema for your service to load in one of two ways:
+1. Using a remote endpoint
+1. Using a local schema file
```
trevor-scheer: 1, 2

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+
+Server projects are configured through a top level `service` key in the config.
+
+```js line=2
+module.exports = {
+  service: { ... },
+};
+```
+
+Setting up a `service` in your Apollo config will save you from needing to use flags your Apollo CLI commands like `apollo service:push` and `apollo service:check`. You can set up the schema for your service to load in one of two ways:
```
trevor-scheer: Personal opinion, I'd rather us not refer to flags at all in the config reference. Flags should be functional and expected behavior, but we should treat the config as _the_ way and not introduce any possible confusion or other ways to do things.

>Defining a `service` key in your Apollo config will provide the CLI with the information it needs to perform commands like `service:push` and `service:check`.

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+1. Using a remote endpoint
+1. Using a local schema file
+
+<h4 id="service-remote-endpoint">Option 1: Remote endpoint</h4>
```
trevor-scheer: Does this give us the functionality to link to specific sections? If so I'd love for us to use these in the client sections as well!

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+      },
+      skipSSLValidation: true, // optional, disables SSL validation check
+    },
+  },
+};
+```
+
+<h4 id="service-local-file">Option 2: Local schema</h4>
+
+In some cases, teams may have a local generated file of their schema that they want to link. This can be either a `.graphql` file with the schma in SDL, or a saved introspection result in `.json`. To link a service with a local file, configure the project like this:
```
trevor-scheer: Going to recommend the same language change as before in the other local schema section.

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+  client: { ... },
+};
+```
+
+### `service`
+
+**Required** –– the CLI and VS Code extension rely on knowledge of your schema to show you "intellisense" (eg. autocomplete on fields, metrics annotations, query validation).
+
+There are a few different ways you can link your client to a schema:
+1. Use the Apollo [schema registry](/docs/platform/schema-registry.html)
```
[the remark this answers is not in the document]
daniman: I actually like using 1. 1. 1. :) I thought it was weird at first, but markdown turns it into a magic normal numbered list, and leaving it like this in the code allows the lines to be _really_ easily rerorrdered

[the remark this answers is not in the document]
trevor-scheer: TIL!

docs/source/references/apollo-config.md
trevor-scheer: Resolved per previous comment

docs/source/references/apollo-config.md
```diff
@@ -0,0 +1,201 @@
...
+```js line=3
+module.exports = {
+  client: {
+    service: 'my-apollo-service@beta', // "beta" is the schema tag we're using
+  },
+};
+```
+If a schema tag is not specified, the default value is `current`.
+
+#### Option 2: Link a schema from a remote endpoint
```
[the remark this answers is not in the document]
daniman: Goooood catch 😸

docs/source/references/apollo-config.md
daniman: I think since there is a default it's okay to mark this as _optional_, but we should document the default behavior fully and improve the error messages :)

daniman:
Great points @trevor-scheer, thanks for the feedback and thorough read!

docs/source/references/apollo-config.md
daniman: Maybe we should have Apollo Client read that preference from the config directly if it exists?

docs/source/references/apollo-config.md
trevor-scheer: Sounds good to me. Looks like I already have an issue open for myself 👍
https://github.com/apollographql/apollo-tooling/issues/702

docs/source/references/apollo-config.md
daniman: Love it, will do 👍

docs/source/references/apollo-config.md
daniman: Yes! Sort of :) `h1`, `h2`, and `h3` all automatically get anchor links in our hexo theme. I was just copying this from what it was before, but it allows us to make that option an anchor link if we want it to be, and to specify exactly what the anchor text will be

PR #286, opened 2018-12-14
Fix example of Local schema

playerx:
First I've tried with sample code provided on Apollo page and I was getting error like this:
> Only absolute URLs are supported

I thought problem was in url, but later I find out that problem was in wrong property name, on another pages there is `url` property but here was `endpoint`


evans:
@playerx Thank you so much!

[additional docs for reference](https://www.apollographql.com/docs/references/apollo-config.html#client-service)

PR #288, opened 2018-12-16
Update schema.md

smolinari:
This updates the article to articulate the understanding that GraphQL should model business logic "the business graph" and isn't just a CRUD system.

As the "[Thinking in Graphs](https://graphql.org/learn/thinking-in-graphs/
)" notes in big letters at the start....

> With GraphQL, you model your business domain as a graph

We should constantly make an effort to make this clear, so the true power of GraphQL, for those who don't "get it" yet, will understand that GraphQL is more than just a CRUD system. 

I'll be going through the whole tutorial with this thinking and will attempt to help in the code too, where possible. 

I hope I can help. 

Scott

### DIFF
--- a/docs/source/tutorial/schema.md
+++ b/docs/source/tutorial/schema.md
@@ -3,11 +3,11 @@ title: "1. Build a schema"
 description: Create a blueprint for your graph's data
 ---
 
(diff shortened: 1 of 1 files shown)


JakeDawkins:
@smolinari Thanks for taking a look at this!

I think some of this could potentially add confusion, so I'm going to close this (also, there hasn't been much conversation or other interest in this).

While there is often quite a lot of business logic within the GraphQL layer (and ideally that logic would grow over time), not all use cases implement it there. Often, people use the GraphQL layer as a means of aggregating data from other data sources.

PR #289, opened 2018-12-16
Update data-source.md

smolinari:
Some small additional updates.

Scott


evans:
@smolinari Thank you so much!

PR #290, opened 2018-12-16
Update resolvers.md

smolinari:
Some additions. 

Scott


evans:
@smolinari Thank you so much!

PR #294, opened 2018-12-27
Improve tutorial docs (data-source page)

cdimitroulas:
Hi there,

I found that many of the variable names were shortened to a single letter or were just shortened versions of a real word. Although this may be a common practice in many codebases (though I disagree with it personally but that's a separate discussion), I feel that tutorial documentation benefits from as much clarity as is practical.

In my PR I have modified some of the variable names which I hope improves the clarity of this page.

I have also made a small code change from `response && response.length` to `Array.isArray(response)` as I believe this again improves the clarity of the tutorial.

I understand many of these choices are very subjective so please feel free to comment and suggest changes as you see fit!

And finally, I hope you are all enjoying yourselves during these Christmas holidays!

### DIFF
--- a/docs/source/tutorial/data-source.md
+++ b/docs/source/tutorial/data-source.md
@@ -46,12 +46,14 @@ _src/datasources/launch.js_
 
 ```js
 async getAllLaunches() {
-  const res = await this.get('launches');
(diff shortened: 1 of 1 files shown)


evans:
@cdimitroulas Absolutely agree! Thank you so much 🎉 

Hope your holidays went well!

PR #296, opened 2018-12-30
Update queries.md

smolinari:
Hi,

Either I was in some lala land and wasn't following this tutorial properly at first, or someone did some changes (which I haven't looked to see if is true or not). At any rate. It's not in the bad shape I thought it was. So, the updates are only minimal.

Scott


evans:
@smolinari Woot! Thank you

PR #304, opened 2019-01-17
Fix getAllLaunches Variable Mismatch

alaneparisot:
Replace `result` with `response`.

### DIFF
--- a/docs/source/tutorial/data-source.md
+++ b/docs/source/tutorial/data-source.md
@@ -49,7 +49,7 @@ _src/datasources/launch.js_
 ```js
 async getAllLaunches() {
   const response = await this.get('launches');
-  return Array.isArray(result)
+  return Array.isArray(response)
     ? response.map(launch => this.launchReducer(launch))
     : [];
 }


docs/source/tutorial/data-source.md
```diff
@@ -49,7 +49,7 @@ _src/datasources/launch.js_
 ```js
 async getAllLaunches() {
   const response = await this.get('launches');
-  return Array.isArray(result)
+  return Array.isArray(response)
```
abernix: I think both of these variables actually need to be changed to `res` in order to match up with [the `fullstack-tutorial` repository's](https://github.com/apollographql/fullstack-tutorial) "final" state, as seen in [this commit](https://github.com/apollographql/fullstack-tutorial/blob/7948b85d747300fd4d1596b6173b9ee783431c75/final/server/src/datasources/launch.js#L28-L33).

docs/source/tutorial/data-source.md
```diff
@@ -49,7 +49,7 @@ _src/datasources/launch.js_
 ```js
 async getAllLaunches() {
   const response = await this.get('launches');
```
abernix:
```suggestion
  const res = await this.get('launches');
```

docs/source/tutorial/data-source.md
```diff
@@ -49,7 +49,7 @@ _src/datasources/launch.js_
...
-  return Array.isArray(result)
+  return Array.isArray(response)
     ? response.map(launch => this.launchReducer(launch))
```
abernix:
```suggestion
    ? res.map(launch => this.launchReducer(launch))
```

docs/source/tutorial/data-source.md
abernix:
```suggestion
  return Array.isArray(res)
```

evans: Let's continue with `response` and change the final state of the `fullstack-tutorial`. I think #294 makes a very good point about `res` vs `response`

abernix: I also agree with, @cdimitroulas!

Let’s make sure we update the tutorial!

cdimitroulas: yay for clearer names :hugs:

PR #310, opened 2019-01-20
Fix redundant triple-quotes in client.md

Enrico2:
The extra triple-quotes caused a formatting error where some non-code block was displayed as code.


abernix:
This was already fixed by #315, which I just merged, but you deserve commit credit too as yours certainly came in first! :)

Thank you!

PR #312, opened 2019-01-21
Fix duplicated heading id in the docs.

jonatanklosko:
The modified header has the exact same `id` as the one above, likely a leftover from copy-pasting =)


abernix:
Thank you!

PR #313, opened 2019-01-21
Pass in the correct variable name

jonrcrowell:
We define the response variable but then pass in an undefined "result" variable


abernix:
Thank you!

PR #316, opened 2019-01-23
Use `yarn` instead of `npm install`

borekb:
The demo project currently contains `yarn.lock`, not `package-lock.json`:

https://github.com/apollographql/fullstack-tutorial/tree/7948b85d747300fd4d1596b6173b9ee783431c75/start/client


borekb:
Thinking about it, it's a bit weird that parts of the tutorial use `npm` while others `yarn` so maybe a more proper solution would be to unify that but for now, the fix is just about running the right command.

peggyrayzis:
I'll fix the tutorial app to remove `yarn.lock`, closing this PR since we want to stick with npm

PR #317, opened 2019-01-23
Fix code example

borekb:
- React and ReactDOM must be imported
- ~~Simple `<div />` is enough for now (importing `<Pages />` doesn't work at this stage)~~ See comment below.


borekb:
It won't be that simple – the `<Pages>` component should be imported but quite a few things are broken in the example project, see https://github.com/apollographql/fullstack-tutorial/issues/63.

borekb:
I've updated the example back to `<Pages>` in 4c39ad7, however, it will only be valid after https://github.com/apollographql/fullstack-tutorial/pull/64 is merged.

docs/source/tutorial/client.md
```diff
@@ -129,7 +129,10 @@ Open `src/index.js` and add the following lines of code:
 _src/index.js_
 
 ```js lines=1,4,6
```
abernix: Just flagging that the additional three lines below requires that these line offsets (used for highlighting) will need to be updated.

I _think_ to:

```suggestion
 ```js lines=3,7,9
```

borekb:
Looking at the current rendering, the line numbers don't seem to have any effect anyway? Maybe the syntax is not entirely correct?

screenshot 2019-01-24 at 16 06 36

abernix:
@borekb 	Looks like it's actually supposed to be `js line=3,7,9` (singular `line` instead of plural `lines`)!  Do you mind trying that?

borekb:
Hi @abernix, sorry for the late reply. I fixed the syntax, however, I no longer have the tutorial in my head as it's been a while since I've gone through it. Hopefully the change still makes sense, it has been part of a collection of other changes, see https://github.com/apollographql/apollo/issues/329.

peggyrayzis:
Closing this PR, we still need to reference the `<Pages/>` component in order for this example to make sense

PR #318, opened 2019-01-23
Fixed code example in Mutations tutorial

borekb:
Removed `resolvers` and `typeDefs` from ApolloClient constructor, they are not needed at this stage yet and have not been imported by any previous example.


JakeDawkins:
This PR would also need to add the changes in the next section to make sure the resolvers and typedefs are introduced back to the constructor

PR #320, opened 2019-01-23
Improved heading structure in tutorial part 7

borekb:
I think that the updated headings should be on the same level as `h2`s on the page, this nesting doesn't feel correct to me:

image


JakeDawkins:
Thanks for opening this @borekb!

I think with the new docs design, this doesn't look quite as bad, since H3's are not shown on the sidebar. I do think that conceptually, the two H3's _should_ fall under the heading of `Update data with Mutation` though, so I'm going to close this one.

PR #321, opened 2019-01-23
[AT-5] Updated VSCode editor extension docs

JakeDawkins:
## Updated editor extension docs

sidebar


docs/source/platform/editor-plugins.md
```diff
@@ -94,16 +96,61 @@ module.exports = {
...
+
+<h3 id="syntax">Syntax highlighting</h3>
+
+Apollo's editor extension provides syntax highlighting for all things GraphQL, including schema definitions in `.graphql` files, complex queries in TypeScript, and even client-only schema extensions. Syntax highlighting for GraphQL works out-of-the-box for `.graphql`, `.gql`, `.js` and `.ts` file types!
+
+<h3 id="autocomplete">Intelligent autocomplete</h3>
+
+Once configured, editors have full knowledge of the schema clients are running operations against, including client-only schemas (for things like local state mutations). Because of this, editors have the ability to autocomplete fields and arguments as you type.
+
+<img src="../images/editors/autocomplete.gif" width="80%" style="margin: 5%" alt="vscode completing a field when typing">
```
peggyrayzis: Something about the line highlighting with the intellisense tooltip is making this gif look janky? Does it look like this with another theme?

docs/source/platform/editor-plugins.md
```diff
@@ -94,16 +96,61 @@ module.exports = {
...
+
+<h3 id="field-type-info">Inline field type information</h3>
+
+Because of GraphQL's strongly-typed schema, editors not only know about which fields and arguments are valid, but also what types are expected. Hover over any type in a valid GraphQL operation to see what type that field returns and whether or not it can be null.
+
+<img src="../images/editors/type-info.png" width="80%" style="margin: 5%" alt="a tooltip showing a Boolean type for a field">
+
+<h3 id="performance-insights">Performance insights</h3>
 
 GraphQL operations provide incredible flexibilty in what data is requested from a service. This can sometimes lead to unknown performance characteristics of how an operation will run. Thanks to the trace wharehouse in the Apollo GraphQL Platform, teams no longer will be surprised by how long an operation takes.
```
peggyrayzis:
```suggestion
GraphQL's flexibility can make it difficult to predict the cost of an operation. Without insight into how expensive an operation is, developers can accidentally write queries that place strain on their graph API's underlying backends. Thanks to the Apollo platform's integration with VS Code and our trace warehouse, teams can avoid these performance issues altogether by instantly seeing the cost of a query right in their editor.

To turn on tracing for your GraphQL server, please visit our [guide](./setup-analytics.html).
```

docs/source/platform/editor-plugins.md
```diff
@@ -94,16 +96,61 @@ module.exports = {
...
-<h2 id="commands">Apollo commands</h2>
+<img src="../images/editors/perf-annotation.png" width="80%" style="margin: 5%" alt="Performance annotation next to a field">
+
+<h3 id="navigating-projects">Navigating projects</h3>
+
+Navigating large codebases can be difficult, but the Apollo GraphQL extension makes this easier than ever. Right-clicking on any field in operations or schemas gives you the ability to jump to (or peek at) definitions, as well as find any other references to that field in your project. Searching a project for any occurrence of a certain field is now a thing of the past!
+
+<img src="../images/editors/jump-to-def.gif" width="80%" style="margin: 5%" alt="Using jump to definition on a fragment">
+
+<h3 id="commands">Apollo commands</h3>
```
peggyrayzis: Switching schema tags should be its own section. The Apollo command to reload the schema should be mentioned in the troubleshooting section.

docs/source/platform/editor-plugins.md
```diff
@@ -3,24 +3,26 @@ title: Editor plugins
...
+- [See performance information](#performance-insights) inline with your query definitions
+- Load GraphQL schemas and queries automatically from an Apollo Config file
+- Add [syntax highlighting](#syntax) for GraphQL files and gql templates inside JavaScript files
+- Detect and load client-side schemas and validates client side field usage in operations
+- [Navigate projects](#navigating-projects) easier with jump-to and peek-at definitions and more
+- Manage [local](#local-schemas) and [client-only](#client-only-schemas) schemas
+- [Switch schema tags](#commands) to work on upcoming features
+- And more...
+
+To get started, first **install the Apollo GraphQL extension** by using this [link](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo) or by searching `Apollo` in the VS Code extension marketplace. After installation, GraphQL syntax highlighting should automatically be enabled for `.graphql`, `.gql`, `.js` and `.ts` file types.
```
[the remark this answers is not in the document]
JakeDawkins: @peggyrayzis This seems like something that should live somewhere else, as it's not really an editor-specific need. Do we already have something like this we could link out to?

docs/source/platform/editor-plugins.md
JakeDawkins: It looks like it's just this theme that does that. Is it bad enough to redo all the screenshots and gifs though? having one with a different theme would be odd

JakeDawkins:
@peggy how would you feel about removing the setup altogether and just pointing to the [configuring Apollo projects](https://www.apollographql.com/docs/references/apollo-config.html#service-config) page, since it pretty much covers everything.

peggyrayzis:
> @peggyrayzis how would you feel about removing the setup altogether and just pointing to the [configuring Apollo projects](https://www.apollographql.com/docs/references/apollo-config.html#service-config) page, since it pretty much covers everything.

I think it's a bad experience to kick the user off to another page that gives them too much information. Until we have the dropdown, I think we should include basic instructions for setting up the configuration via the schema registry and link to the config page for more advanced options. Right now, we're still telling them that the first step should be to install the extension, which is going to lead to a bad experience since most of the features we're advertising won't work.

peggyrayzis:
@jbaxleyiii @JakeDawkins I'm going to merge and approve, because I think we should ship these today with the new AC 2.5 release. There still might be some changes I want to make later, but I don't think they're important enough to block.

PR #308, opened 2019-01-19
Fix typo in variable name (result -> response)

Enrico2:

PR #322, opened 2019-01-23
Update data fetch method - "result" undefined error

tsettle44:
Array.isArray(result) -- "result" is undefined
Array.isArray(response) -- correct


abernix:
Duplicate of #304 and #308.  Thank you for opening this though!

PR #323, opened 2019-01-23
Fix code example in local state

borekb:
Added `StyledButton` so that `LogoutButton` works.


JakeDawkins:
Thanks for this @borekb!

PR #325, opened 2019-01-24
Data source tutorial: Fix var name in getAllLaunches()

TheDom-Test:


TheDom-Test:
Posted from wrong account 😅

PR #291, opened 2018-12-16
Update production.md Fixes (sort of) #239

smolinari:
Some corrections. Sort of fixes #239. 

Scott

PR #327, opened 2019-01-24
Rework service:check explanation

trevor-scheer:
Recommend config first, and provide more prominent links out to the config docs.

### DIFF
--- a/docs/source/platform/schema-validation.md
+++ b/docs/source/platform/schema-validation.md
@@ -131,15 +131,37 @@ Of course, it’s also possible to leave the field in place indefinitely!
 
 <h2 id="cli">Checking schema changes with the Apollo CLI</h2>
 
-To check and see the difference between the current published schema and a new version, run the following command, substituting the appropriate GraphQL endpoint URL and an API key:
+To check and see the difference between the current published schema and a new version, make sure you have a valid [service configuration](../resources/apollo-config.html#service-config) like so, substituting the appropriate GraphQL endpoint URL and an API key:
+
+```bash
+# .env
+ENGINE_API_KEY=<API_KEY>
+```
 
 > An API key can be obtained from a service's _Settings_ menu within the [Apollo Engine dashboard](https://engine.apollographql.com/).
 
-```bash
-apollo service:check --key="<API_KEY>" --endpoint="http://localhost:4000/graphql"
+```javascript
+// apollo.config.js
+module.exports = {
+  service: {
+    endpoint: {
+      url: 'http://localhost:4000/graphql'
+    }
+  }
+};
 ```
 
-> For accuracy, it's best to retrieve the schema from a running GraphQL server (with introspection enabled), though the CLI also reference a local file. See [config options](../platform/apollo-config.html) for more information.
+And run the following command:
+
+```bash
+npx apollo service:check
+```
+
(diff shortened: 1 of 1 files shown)


docs/source/platform/schema-validation.md
```diff
@@ -131,15 +131,37 @@ Of course, it’s also possible to leave the field in place indefinitely!
 
 <h2 id="cli">Checking schema changes with the Apollo CLI</h2>
 
-To check and see the difference between the current published schema and a new version, run the following command, substituting the appropriate GraphQL endpoint URL and an API key:
+To check and see the difference between the current published schema and a new version, make sure you have a valid [service configuration](../resources/apollo-config.html#service-config) like so, substituting the appropriate GraphQL endpoint URL and an API key:
+
+```bash
+# .env
+ENGINE_API_KEY=<API_KEY>
```
evans: I think we need to be more explicit that this line goes inside of an `.env` file and where the `.env` file should be saved. Additionally it can be confusing what the api key is, something like [this](https://github.com/apollographql/apollo/pull/291/files#diff-c3154587c474c91ac2bac208f0ed3accL20) helps a lot

docs/source/platform/schema-validation.md
```diff
@@ -131,15 +131,37 @@ Of course, it’s also possible to leave the field in place indefinitely!
...
+
 ```bash
-apollo service:check --key="<API_KEY>" --endpoint="http://localhost:4000/graphql"
+npx apollo service:check
 ```
 
-> For accuracy, it's best to retrieve the schema from a running GraphQL server (with introspection enabled), though the CLI also reference a local file. See [config options](../platform/apollo-config.html) for more information.
+> Alternatively, you may skip configuration and pass flags to the command, however we recommend [configuring](../resources/apollo-config.html) the project. The configuration file behaves as a source of truth for various Apollo features including IDE support and other CLI commands.
+
+```bash
```
evans: https://deploy-preview-327--apollo-docs-index.netlify.com/docs/platform/schema-validation.html#cli gets a bit busy and hard to follow which portion/code is an aside. Can we make this code snippet quoted or place it after the rest of this section? Alternatively we could add a small header before the next paragraph

You can run the docs locally with `npm start` and they auto update thanks to @abernix's additions

docs/source/platform/schema-validation.md
```diff
@@ -131,15 +131,37 @@ Of course, it’s also possible to leave the field in place indefinitely!
 
 <h2 id="cli">Checking schema changes with the Apollo CLI</h2>
 
-To check and see the difference between the current published schema and a new version, run the following command, substituting the appropriate GraphQL endpoint URL and an API key:
+To check and see the difference between the current published schema and a new version, make sure you have a valid [service configuration](../resources/apollo-config.html#service-config) like so, substituting the appropriate GraphQL endpoint URL and an API key:
```
evans: Reading through the doc now, this lead-in sentence is surprising with the preceding header(the original content definitely pushed in this direction just happened to notice with your change 😄). In my mind, there is something missing: an indication that this is a setup section in the title or an explanation of the check/cli's utility or both. 

The explanation of the cli's effect at the end of this section would probably provide a nice motivation with some rewording before diving into the setup, maybe indicated with a subheader.

daniman:
@trevor-scheer I made some substantial modifications to the schema-validation.md doc over the last few weeks in line with the latest schema validation work. I think that has unfortunately introduced some very hard to reconcile merge conflicts. Let me know how you would like to proceed with this PR in particular.

PR #328, opened 2019-01-24
Fixed error in getAllLaunches()

justinyum98:
In the method getAllLaunches(), Array.isArray(result) should be Array.isArray(response).


evans:
@justinyum98 Thank you so much! We've merged #308

PR #330, opened 2019-01-25
typo: replace result with response in code snippet

nanu146:


daniman:
Thanks ver  much @nanu146! I actually just merged #331 with the exact same fix, but you got there first, I just didn't see yours first 😢

PR #331, opened 2019-01-25
Fix Tutorial Typo

kylekarpack:
Fixed getAllLaunches method to reference "response" instead of a "result" variable that does not exist


daniman:
Thanks for the fix @kylekarpack!

PR #333, opened 2019-01-29
(op-reg) Update for `forbidUnregisteredOperations` predicate function support.

abernix:
The Apollo Server operation registry plugin — available for Apollo Engine
Team and Enterprise users, will SOON (in the next day or so) support passing a predicate function rather than a boolean on the `forbidUnregisteredOperations` parameter.

This new parameter can be used to selectively enforce the safe-listing
depending on contextual differences — for example, the presence of a header.

This documentation adds the implementation details for that new feature
along with an example of how to utilize it.

### DIFF
--- a/docs/source/platform/operation-registry.md
+++ b/docs/source/platform/operation-registry.md
@@ -188,6 +188,51 @@ Execution forbidden
 
 Finally, to confirm that the server will allow permitted operations, try running an operation from the client.
 
+## Configuration
+
+### Selective enforcement
+
+In some cases, deployments may want to selectively enable the behavior of `forbidUnregisteredOperations` depending on environmental conditions (e.g. based on headers).
+
+To selectively enable operation safe-listing, the `forbidUnregisteredOperations` setting supports a predicate function which receives the request context and can return `true` or `false` to indicate whether enforcement is enabled or disabled respectively.
+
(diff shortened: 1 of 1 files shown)


abernix:
# ⚠️ 

This isn't released yet, so please don't merge _just_ yet!

abernix:
Ok, this is now in `apollo-server-plugin-operation-registry@latest` so this is no longer blocked on the release! 🎉

docs/source/platform/operation-registry.md
```diff
@@ -188,6 +188,51 @@ Execution forbidden
...
+      // request `context`.  Additional validation is possible, but this
+      // function must be synchronous.  For more details, see the note below.
+      forbidUnregisteredOperations({
+        context, // Destructure the shared request `context`.
+        request: {
+          http: { headers } // Destructure the `headers` class.
+        }
+      }) {
+        // If a magic header is in place, allow any unregistered operation.
+        // The `headers` implements the Fetch API which is documented here:
```
peggyrayzis: I would move the sentence that headers implements the Fetch API up to the main content since it's an important detail.

abernix: Thank you, and done via https://github.com/apollographql/apollo/pull/333/commits/c00b07ae49635f71a9243c4d7acc30ff8108141c.

PR #334, opened 2019-01-29
(op-reg) General cleanup of the operation registry documentation.

abernix:
I'll let the change-set speak for itself!

### DIFF
--- a/docs/source/platform/operation-registry.md
+++ b/docs/source/platform/operation-registry.md
@@ -5,6 +5,8 @@ description: How to secure your graph with operation safelisting
 
 ## Overview
 
+> The operation registry is an Apollo Platform feature available on the [_Team_ and _Enterprise_ plans](https://www.apollographql.com/plans/).  To get started with the Apollo Platform, begin with [the documentation](https://www.apollographql.com/docs/).
+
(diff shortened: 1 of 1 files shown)


docs/source/platform/operation-registry.md
```diff
@@ -24,21 +26,28 @@ Operations defined within client applications are automatically extracted and up
...
 
 ```
 npm install apollo --save-dev
 ```
 
-> Yarn users can run `yarn add apollo --dev`.
+> Yarn users should run `yarn add apollo --dev`.
 
-**2. Register the server's schema with Apollo:**
+**2. Register the server's schema with Apollo Engine:**
```
peggyrayzis:
```suggestion
**2. Push your schema to the Apollo registry:**
```

I try to shy away from describing the schema registry as a feature of Engine because many people associate Engine with our paid plans even though the schema registry is free. Additionally, I think associating it with Engine instead of the platform goes against what we're trying to achieve with the overall platform story. What do you think?

abernix: @peggyrayzis Could we use your suggestion, but still call it the `schema registry` (we have two registries, and I think it's worth clarifying which one we're talking about, specifically in this article which touches on both)?  E.g.:

```suggestion
**2. Push your schema to the Apollo schema registry:**
```

peggyrayzis: I like that suggestion!

PR #339, opened 2019-02-04
Add client version to operation registration docs

justinanastos:


docs/source/platform/operation-registry.md
```diff
@@ -66,26 +66,47 @@ abc123  <service>     current
 
 > If you encounter any errors, refer to the _**Troubleshooting**_ section below.
 
-**3. Register operations from the client bundle.**
+**3. Configure client reporting in the client bundle.**
 
-Now we'll use `apollo client:push` to locate operations within the client codebase and upload a manifest of those operations to Apollo operation registry. Once Apollo Server has been configured to respect the operation registry, only operations which have been included in the manifest will be permitted.
+The client must identity itself so we can match operations with a client and version. See the [Identifying clients](../platform/client-awareness.html#Setup) documentation to configure `ApolloClient` to identity itself. The following is an example:
```
JakeDawkins: I don't think clients _must_ identify themselves to use the operation registry. It's a nice-to-have, but not required from my understanding

docs/source/platform/operation-registry.md
```diff
@@ -66,26 +66,47 @@ abc123  <service>     current
...
+    uri: 'http://localhost:4000/graphql',
+  }),
+  name: 'insert your client name',
+  version: 'insert your client version',
+});
+```
+
+**4. Register operations from the client bundle.**
+
+Now we'll use [`apollo client:push`](https://github.com/apollographql/apollo-tooling#apollo-clientpush) to locate operations within the client codebase and upload a manifest of those operations for a specific [client version](../platform/client-awareness.html) to the Apollo operation registry. Once Apollo Server has been configured to respect the operation registry, only operations which have been included in the manifest will be permitted for the corresponding client version.
```
JakeDawkins: Same as above. I think registrations aren't client-aware. The client-awareness of operations only comes when the client _makes_ requests, and engine matches those clients with the registry.

justinanastos: After talking it over, @JakeDawkins is totally right. Client version is _not_ necessary, just a nice-to-have.

justinanastos:
As @JakeDawkins pointed out, the client version is _not_ required. Therefore, these docs don't need this change.

PR #340, opened 2019-02-05
Update schema.md

cfjedimaster:
I felt like the text saying to include schema stuff "inside the gql function" may not be quite clear enough for folks who haven't seen template strings or perhaps haven't seen template strings used as a function in that style.


daniman:
Thank you for the PR @cfjedimaster!

PR #341, opened 2019-02-07
Update `service:check` docs with new flags

trevor-scheer:
This PR adds documentation for 3 new flags for the `service:check` CLI command:

* validationPeriod
* queryCountThreshold
* queryCountThresholdPercentage

Also removes a broken link to performance metrics `./performance.html`

This PR is best read by commits, since the second commit is strictly prettier re-formatting the markdown file.

### DIFF
--- a/docs/source/platform/schema-validation.md
+++ b/docs/source/platform/schema-validation.md
@@ -151,7 +151,7 @@ After analyzing the changes against current usage metrics, Apollo will identify 
 By default, the CLI will look at operations run within the last day in order to decide which changes are breaking. This period of time may not be large enough, so the `validationPeriod` flag accommodates custom timeframes:
 
 ```bash
-apollo service:check --validationPeriod="P2W"
+apollo service:check --validationPeriod=P2W
 ```
 
 > Valid durations are represented in ISO 8601. For reference, see: https://bit.ly/2DEJ3UN. It can also be provided as a number in seconds, i.e. 86400 for a single day.
--- a/docs/source/platform/schema-validation.md
+++ b/docs/source/platform/schema-validation.md
@@ -162,6 +162,8 @@ Two other parameters for customizing the results of `service:check` are threshol
 - `queryCountThresholdPercentage` - Similar to `queryCountThreshold`, but expressed as a percentage of all operation volume.
   > Note: these flags are compatible with each other. In the case that both are provided, an operation must meet or exceed both thresholds.
 
+If you have requests for other filtering or threshold mechanisms, we'd love to hear them! Please feel free to submit a [feature request](https://github.com/apollographql/apollo-tooling/issues/new?template=feature-request.md) or PR to the [apollo-tooling](https://github.com/apollographql/apollo-tooling/) repo.
+
 ```bash
 apollo service:check \
 # Validate the schema against operations that have run in the last 5 days
--- a/docs/source/platform/schema-validation.md
+++ b/docs/source/platform/schema-validation.md
@@ -19,7 +19,7 @@ It might be tempting to version a GraphQL API the same way, but it's unnecessary
 
 <h3 id="field-usage">Field usage</h3>
 
-Rather than returning extensive amounts of data which might not be necessary, GraphQL allows consumers to specify exactly what data they need. This field-based granularity is valuable and avoids "over-fetching" but also makes it more difficult to understand what data is currently being used.
+Rather than returning extensive amounts of data which might not be necessary, GraphQL allows consumers to specify exactly what data they need. This field-based granularity is valuable and avoids "over-fetching", but also makes it more difficult to understand which parts of the schema are being used.
 
 To improve the understanding of field usage within an API, Apollo Server extends GraphQL with rich tracing data that demonstrates _how_ a GraphQL field is used and _when_ it's safe to change or eliminate a field.
 
--- a/docs/source/platform/schema-validation.md
+++ b/docs/source/platform/schema-validation.md
@@ -158,8 +158,9 @@ apollo service:check --validationPeriod="P2W"
 
 Two other parameters for customizing the results of `service:check` are threshold values. For example, you may wish to drop support for an old version of an app in order to remove some deprecated fields. Using these parameters, you can decide what amount of breakage is acceptable before shipping any breaking changes.
 
-- `queryCountThreshold` - This flag will only validate the schema against operations that have been executed the specified number of times within the provided duration.
-- `queryCountThresholdPercentage` - Similar to `queryCountThreshold`, but expressed as a percentage of all operations.
+- `queryCountThreshold` - This flag will only validate the schema against operations that have been executed at least the specified number of times within the provided duration.
+- `queryCountThresholdPercentage` - Similar to `queryCountThreshold`, but expressed as a percentage of all operation volume.
+  > Note: these flags are compatible with each other. In the case that both are provided, an operation must meet or exceed both thresholds.
 
 ```bash
 apollo service:check \
--- a/docs/source/platform/schema-validation.md
+++ b/docs/source/platform/schema-validation.md
@@ -154,7 +154,7 @@ By default, the CLI will look at operations run within the last day in order to 
 apollo service:check --validationPeriod=P2W
 ```
 
-> Valid durations are represented in ISO 8601. For reference, see: https://bit.ly/2DEJ3UN. It can also be provided as a number in seconds, i.e. 86400 for a single day.
+> Valid durations are represented in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations). It can also be provided as a number in seconds, i.e. 86400 for a single day.
 
 Two other parameters for customizing the results of `service:check` are threshold values. For example, you may wish to drop support for an old version of an app in order to remove some deprecated fields. Using these parameters, you can decide what amount of breakage is acceptable before shipping any breaking changes.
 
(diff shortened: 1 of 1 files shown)


trevor-scheer:
Really open to feedback on this. I'd love to hear a better way to explain a real-life use case for these, while still keeping the explanation concise.

JakeDawkins:
I think the examples at the end makes these docs make perfect sense :)

zionts:
Sorry, I missed a top-level comment, but there are just a few little tweaks and tidbits I'd like to understand before a merge :)

docs/source/platform/schema-validation.md
```diff
@@ -34,13 +33,13 @@ We'll go over these two kinds of field rollovers separately and show how to make
 
 <h3 id="renaming-or-removing">Renaming or removing a field</h3>
 
-When a field is unused, renaming or removing it is as straightforward as it sounds: it can be renamed or removed.  Unfortunately, if a GraphQL deployment doesn't have per-field usage metrics, additional considerations should be made.
+When a field is unused, renaming or removing it is as straightforward as it sounds: it can be renamed or removed. Unfortunately, if a GraphQL deployment doesn't have per-field usage metrics, additional considerations should be made.
```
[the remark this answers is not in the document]
trevor-scheer: Resolved in e72dffa672ebe207018466c22a5d75855543988d

docs/source/platform/schema-validation.md
```diff
@@ -147,7 +146,30 @@ After analyzing the changes against current usage metrics, Apollo will identify
...
+```bash
+apollo service:check --validationPeriod="P2W"
+```
+
+> Valid durations are represented in ISO 8601. For reference, see: https://bit.ly/2DEJ3UN. It can also be provided as a number in seconds, i.e. 86400 for a single day.
+
+Two other parameters for customizing the results of `service:check` are threshold values. For example, you may wish to drop support for an old version of an app in order to remove some deprecated fields. Using these parameters, you can decide what amount of breakage is acceptable before shipping any breaking changes.
+
+- `queryCountThreshold` - This flag will only validate the schema against operations that have been executed the specified number of times within the provided duration.
+- `queryCountThresholdPercentage` - Similar to `queryCountThreshold`, but expressed as a percentage of all operations.
```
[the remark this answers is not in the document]
trevor-scheer: Resolved in 5375cb6ed7daa8fc54dd22bc0fc7d043e7d8be68

docs/source/platform/schema-validation.md
```diff
@@ -147,7 +146,30 @@ After analyzing the changes against current usage metrics, Apollo will identify
...
-The more [performance metrics](./performance.html) that Apollo has, the better the report of these changes will become.
+<h3 id="cli-advanced">Advanced CLI Usage</h3>
+
+By default, the CLI will look at operations run within the last day in order to decide which changes are breaking. This period of time may not be large enough, so the `validationPeriod` flag accommodates custom timeframes:
+
+```bash
+apollo service:check --validationPeriod="P2W"
+```
+
+> Valid durations are represented in ISO 8601. For reference, see: https://bit.ly/2DEJ3UN. It can also be provided as a number in seconds, i.e. 86400 for a single day.
```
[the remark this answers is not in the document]
trevor-scheer: So the quotes aren't needed, these will all be received as a string. I went ahead and removed the quotes around P2W for consistency, but I'm not opposed to putting them everywhere. I don't know what's standard or best practice.

docs/source/platform/schema-validation.md
```diff
@@ -9,18 +9,17 @@ A GraphQL schema can change in a number of ways between releases and, depending
...
+It might be tempting to version a GraphQL API the same way, but it's unnecessary with the right techniques. By following the strategies and precautions outlined in this guide and using Apollo tooling that adds clarity to every change, many iterations of an API can be served from a single endpoint.
 
 <h3 id="field-usage">Field usage</h3>
 
-Rather than returning extensive amounts of data which might not be necessary, GraphQL allows consumers to specify exactly what data they need.  This field-based granularity is valuable and avoids "over-fetching" but also makes it more difficult to understand what data is currently being used.
+Rather than returning extensive amounts of data which might not be necessary, GraphQL allows consumers to specify exactly what data they need. This field-based granularity is valuable and avoids "over-fetching" but also makes it more difficult to understand what data is currently being used.
```
zionts: small nit: It's not about "what data is currently being used" it's about "what part of the schema are actually being used"
The data might be used elsewhere.

As I said, very tiny nit.

docs/source/platform/schema-validation.md
```diff
@@ -64,8 +63,8 @@ const getUserResolver = (root, args, context) => {
 const resolvers = {
   Query: {
     getUser: getUserResolver,
-    user: getUserResolver,
-  },
+    user: getUserResolver
```
zionts: why are we removing the trailing commas?

docs/source/platform/schema-validation.md
```diff
@@ -147,7 +146,31 @@ After analyzing the changes against current usage metrics, Apollo will identify
 2. **Warning**: There are potential problems that may come from this change, but no clients are immediately impacted.
 3. **Notice**: This change is safe and will not break current clients.
 
-The more [performance metrics](./performance.html) that Apollo has, the better the report of these changes will become.
+<h3 id="cli-advanced">Advanced CLI Usage</h3>
+
+By default, the CLI will look at operations run within the last day in order to decide which changes are breaking. This period of time may not be large enough, so the `validationPeriod` flag accommodates custom timeframes:
```
zionts: huh? Isn't the default to use the maximum time window by the paid-for retention of the plan? If not, then why did we change that? When you don't pass any historical parameters, that's what it does.

zionts: We should say "the `validationPeriod` is always the duration to include prior to **now**" or the like. Let's make it clear that the "to" is always "right now" and you can't override that atm.

docs/source/platform/schema-validation.md
```diff
@@ -147,7 +146,31 @@ After analyzing the changes against current usage metrics, Apollo will identify
...
+
+```bash
+apollo service:check --validationPeriod=P2W
+```
+
+> Valid durations are represented in ISO 8601. For reference, see: https://bit.ly/2DEJ3UN. It can also be provided as a number in seconds, i.e. 86400 for a single day.
```
zionts: Could we put the link in one of these [things](www.things.things.things)?

docs/source/platform/schema-validation.md
```diff
@@ -147,7 +146,31 @@ After analyzing the changes against current usage metrics, Apollo will identify
...
+
+Two other parameters for customizing the results of `service:check` are threshold values. For example, you may wish to drop support for an old version of an app in order to remove some deprecated fields. Using these parameters, you can decide what amount of breakage is acceptable before shipping any breaking changes.
+
+- `queryCountThreshold` - This flag will only validate the schema against operations that have been executed at least the specified number of times within the provided duration.
+- `queryCountThresholdPercentage` - Similar to `queryCountThreshold`, but expressed as a percentage of all operation volume.
+  > Note: these flags are compatible with each other. In the case that both are provided, an operation must meet or exceed both thresholds.
```
zionts: Got it. They're additive, thanks!

docs/source/platform/schema-validation.md
```diff
@@ -147,7 +146,31 @@ After analyzing the changes against current usage metrics, Apollo will identify
 2. **Warning**: There are potential problems that may come from this change, but no clients are immediately impacted.
 3. **Notice**: This change is safe and will not break current clients.
 
-The more [performance metrics](./performance.html) that Apollo has, the better the report of these changes will become.
+<h3 id="cli-advanced">Advanced CLI Usage</h3>
```
zionts: I think that many people reading this will have other features that they'd like to have (e.g. ignoring errors in deprecated fields, ignoring a specific "whitelist" of field changes, modifying the "to" parameter, etc.) and I think it'd be wise to tell them "we know this isn't everything you might want to tune. If you have a specific request, we'd love to hear it. Please open up an issue <here> and provide as much detail as you can"

docs/source/platform/schema-validation.md
trevor-scheer: Prettier!

docs/source/platform/schema-validation.md
evans: The default is a day. We haven't had a chance to implement the default as the max time window yet

docs/source/platform/schema-validation.md
trevor-scheer: Resolved in 14365d7b3e8577b70334d05c5ab31bea4e652bfd

docs/source/platform/schema-validation.md
trevor-scheer: Resovled in 118f491f5b67266c2f89f7da62d65d4b52bd5983

docs/source/platform/schema-validation.md
trevor-scheer: Resolved in edb9ee34211ebc1224c4b58e126bc1c5674f780b

docs/source/platform/schema-validation.md
trevor-scheer: Addressed both here: d86144beab8cb26492104b3204dac64248509315

I opted not to specify about defaults since it varies from plan to plan, but if you see a good way to present that information I'm open to it.

Based on previous behavior and my incorrect assumption, I'll need to update behavior of the CLI with the respective change. This is a really good catch, thank you 😄

PR #343, opened 2019-02-11
[WIP] Describe permanence of operation registry

justinanastos:
This has caused some confusion. I didn't understand that the operation registry is a bucket filled with valid operation hashes that grew forever. I thought we were going to be authorizing specific clients and versions to specific operations, that is not the case.

### DIFF
--- a/docs/source/platform/operation-registry.md
+++ b/docs/source/platform/operation-registry.md
@@ -17,6 +17,8 @@ The Apollo Operation Registry allows organizations to:
 
 Operations defined within client applications are automatically extracted and uploaded to Apollo Engine using the Apollo CLI. Apollo Server fetches a manifest of these operations from Apollo Engine and forbids execution of operations which were not registered from the client bundle.
 
+All operations that are registered will be permanently stored. Future operation uploads to Apollo Engine will not invalidate operations previously registered. You can even upload operations from legacy applications to ensure their service will not be affected by operation safelisting.
+
 ## Getting started
 
 ### Prerequisites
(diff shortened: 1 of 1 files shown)


docs/source/platform/operation-registry.md
```diff
@@ -95,6 +97,8 @@ npx apollo client:push              \
     --includes="src/**/*.{ts,js,graphql}"
 ```
 
+Note that you can use `clientName`, `clientTag`, and `clientVersion`. These fields are retained for analysis, not for validation. All operations that are stored in the operation registry are legal for _all_ clients.
```
abernix: Rather than having this as a separate "Note", could we merge the bit about `clientName`, `clientTag` and `clientVersion` into the bullet-points above, or the example itself (or both)?

abernix: Similarly, for the remaining bit, I think we could move this directly below the bullet-points. It's important!

Also, what if we provided a bit more concrete explanation as to _"why?"_ since, I suspect, not understanding the need for client-side cryptography was a reason this behavior wasn't more clear:

```suggestion
While operation registrations do track _which_ client they were registered from (e.g. `clientName`, `clientVersion`), **all registered operations are permitted from any client** since operations in client bundles aren't _themselves_ cryptographically signed and GraphQL endpoints can receive requests from any client (e.g. `curl`, Insomnia, Postman, etc.).  In the future, in combination with Apollo Engine usage metrics, we suspect this client metadata will assist in removing operations from the safe-list which are no longer used (e.g. deprecated clients).
```

Thoughts?

PR #345, opened 2019-02-15
fix broken APC link

### LINKED ISSUE #344: broken links to APC
All the links I try to find for APC are broken in the docs

they link to https://www.apollographql.com/docs/guides/performance.html#automatic-persisted-queries

oppianmatt:
pointing it to old docs unless something better?

fixes #344


oppianmatt:
I have signed and submitted

zionts:
@oppianmatt , thanks for the PR! I think that https://www.apollographql.com/docs/apollo-server/whats-new.html#Automatic-Persisted-Queries would likely be a better link. Let me know if you agree :)

abernix:
Thanks!  I've updated this to point to the resurrected location of this document, which is in https://www.apollographql.com/docs/apollo-server/features/apq/.

PR #336, opened 2019-01-30
Rework schema publication + validation [BE-67]

evans:
This PR updates the schema registry(publication and validation) pages to include the most recent dogfooding information/user feedback.

Some of the learnings include: running a publish on deploy and optionally for the mainline branch during CI, validating multiple tagged schemas, use `npx` instead of global apollo, and enumerating the categories of changes. Additionally we rework the validation page to have a more compelling/informative introduction that presents setup faster and explain strategies for making changes.

We also remove second person language, swap Apollo Engine for Apollo UI where appropriate, and clean up some parallel construction.

> Note: the strategies for proper schema evolution should probably be expanded upon an placed in a separate guide eventually

PR #346, opened 2019-02-19
Update: Non Breaking Changes documentation

14850842:
Having `groupId` as a required field would be a breaking change. As existing clients would now require this field when running this query.

### DIFF
--- a/docs/source/platform/schema-validation.md
+++ b/docs/source/platform/schema-validation.md
@@ -98,7 +98,7 @@ type Query {
   getUsers(ids: [ID!]!): [User]!
 
   # what we want to end up with
-  getUsers(ids: [ID!], groupId: ID!): [User]!
+  getUsers(ids: [ID!], groupId: ID): [User]!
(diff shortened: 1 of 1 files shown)


docs/source/platform/schema-validation.md
```diff
@@ -98,7 +98,7 @@ type Query {
   getUsers(ids: [ID!]!): [User]!
 
   # what we want to end up with
-  getUsers(ids: [ID!], groupId: ID!): [User]!
+  getUsers(ids: [ID!], groupId: ID): [User]!
```
zionts: Good catch! I think this might get overridden by https://github.com/apollographql/apollo/pull/336 and #327, but happy to ship this for the mean time :) Thank you for the contribution.

daniman:
Thanks for the PR @14850842! Unfortunately we made some updates to this article earlier this week that removed this section, but this is a great catch and we appreciate the contribution :)

PR #347, opened 2019-02-24
Fix typo in Editor plugins docs

lauragift21:


daniman:
Thanks so much @lauragift21!

PR #351, opened 2019-03-03
Add import for React and ReactDOM

billfienberg:
In the _Connect your client to React_ section, the code doesn't work without importing `React` and `ReactDOM`.


daniman:
Thanks @billfienberg!

PR #352, opened 2019-03-04
client-awareness: Change many uses of "api" to "API".

abernix:
An "application programming interface" need not be capitalized in its full form, but I believe API should remain capitalized.

With all due respect to SCUBA, I'm not willing to let every form initialism land in the dictionary in lower-case.


daniman:
Thanks @abernix, LGTM 🚢

PR #353, opened 2019-03-04
client-awareness: Add clarity to "Advanced setup" section.

abernix:
By default, Apollo Server and Apollo Client take care of most aspects of client awareness automatically, with the requirements being:

* Using recent versions of Apollo Server and Apollo Client which support client awareness automatically.
* The `name` and `version` be set on the `ApolloClient` constructor.

As currently written, the variables used in this _Advanced setup_ section have caused some to believe that those are the headers that should be used, whereas they were merely suggestions.  The _Setup_ section above should still be observed for most cases and I hope the additional clarity in this re-wording will help avoid future confusion.

cc  @michael-watson


michael-watson:
This looks good to me!

PR #356, opened 2019-03-05
Update Apollo community from Slack to Spectrum.

atitoa93:
Note: I noticed that the old URL `https://www.apollographql.com/slack` redirects to the Spectrum community, but I'm not sure if it does more than redirecting like analysis or something like that, So, please take that on consideration while reviewing the PR.


evans:
@atitoa93 thank you!

PR #357, opened 2019-03-06
Clarify appropriate location for schema code in tutorial

garrettgsb:
The existing documentation mentions that "_Your schema code will go inside the `gql` function,_" but I found that easy to miss and only informative in retrospect. This change adds another line immediately before the reader needs to know where the SDL code goes to help curb the confusion that I went through.


daniman:
Thanks very much @garrettgsb!

PR #362, opened 2019-03-25
Fix confusing highlight on data-source tutorial

frnk94:
## Problems
The doc highlight can lead to user being confuse about what needed to be added in `src/index.js`

## Caveat
I do agree that Highlighting most of the box seems weird and remove the purpose of the highlight in the first place. But just highlighting a part of it is counter intuitive and can lead to the user being confuse.

### Before
Screen Shot 2019-03-25 at 9 47 34 AM

### After
Screen Shot 2019-03-25 at 9 55 34 AM


daniman:
Thanks @frnk94!

PR #371, opened 2019-03-27
remove `resolvers` & `typeDefs` fields

dborysov:
These two fields should be introduced later in the course


JakeDawkins:
This is a duplicate of #318 :)

PR #381, opened 2019-04-02
Adding forgotten import statement

ObsidianCat:
Instead of my closed pull request (I closed the original one because I have problems to do rebase after resetting from accidental merge).


evans:
@ObsidianCat Thank you so much!

PR #392, opened 2019-04-11
Added missing async keyword

chaitanyagurrapu:


trevor-scheer:
Thanks for the contribution @chaitanyagurrapu!

Adding this `async` is extraneous since we're not `await`ing anything inside of the function. It's ok for functions to return a `Promise` without being marked as `async`.

PR #412, opened 2019-04-16
Add client resolvers in correct step

JakeDawkins:
Previously, the `resolvers`, and `typeDefs` were getting added into the tutorial too early without an import.

This PR removes them from the `mutations` step and adds them back in the `local state` step with a note about imports

resolves #318


trevor-scheer:
LGTM!

PR #427, opened 2019-04-24
Fix protobuf link

tommista:


daniman:
Thanks very much @tommista!

PR #428, opened 2019-04-25
Fix Glossary links, grammar, clarity

lorensr:


daniman:
Awesome, thanks so much for the edits @lorensr!

PR #436, opened 2019-04-28
Update resolvers.md

Ruffeng:
The first code snippet was missing the name of the file is editing. Even though the description explains which file has to be edited, the rest of the code snippets contain the path of the file to modify.

So I added it to make it more hegemony along the rest of the code snippets.


daniman:
Great catch, thanks so much for the PR!

PR #440, opened 2019-05-07
op-reg: Add notes about schema tag limitations.

abernix:
Currently, the operation registry does not support any schema tags besides the default schema tag of `current`.  Users can still use the operation registry, but they should make sure that they also register their schema to the `current` tag.

### DIFF
--- a/docs/source/platform/operation-registry.md
+++ b/docs/source/platform/operation-registry.md
@@ -34,7 +34,7 @@ Operations defined within client applications are automatically extracted and up
 
   To use the operation registry with schema tags, the schema which necessitates demand control should also be registered to the default (`current`) tag for the same service.  For example, if a service is using a `prod` schema tag and publishing the schema with `apollo service:push --tag=prod`, the same schema should also be pushed to the default tag with `apollo service:push --tag=current`.
 
-Please contact the Apollo sales team if any of these limitations are necessary.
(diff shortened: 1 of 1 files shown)


docs/source/platform/operation-registry.md
```diff
@@ -22,15 +22,23 @@ Operations defined within client applications are automatically extracted and up
...
   - To obtain an API key, visit [Apollo Engine](https://engine.apollographql.com) and create a service.
+  
+### Limitations
+
+- Subscriptions within Apollo Server should be disabled.  For more information, see the instructions below.
+- Only the default schema tag (`current`) is supported.
+
+  To use the operation registry with schema tags, the schema which necessitates demand control should also be registered to the default (`current`) tag for the same service.  For example, if a service is using a `prod` schema tag and publishing the schema with `apollo service:push --tag=prod`, the same schema should also be pushed to the default tag with `apollo service:push --tag=current`.
+
+Please contact the Apollo sales team if any of these limitations are necessary.
```
evans: This reads a bit odd to me. I suggest something like `Please contact the Apollo sales team if you require a solution to any of these limitations.`

abernix:
Thank you both, @evans and @trevor-scheer!

Agree about the separate PR for the _Troubleshooting_ changes.  We'll want to get that in soon, but I'll have to try to re-visit another time.

PR #442, opened 2019-05-09
Replaced images containing invalid queries

ioveracker:
The old images referenced a `year` property on the `Launch` type, which no longer exists in the tutorial. It has been replaced with the `site` property.


daniman:
Excellent, thanks so much for the new pics @ioveracker!

Issue #264, opened 2018-11-26
Incomplete instructions when publishing to ZEIT Now


semireg:
The (tutorial)[https://www.apollographql.com/docs/tutorial/production.html#deploy] assumes that a vanilla `npx now` is going to work. My experience is it only publishes a static site because of missing now.json.

ZEIT publishes an (Apollo example)[https://github.com/zeit/now-examples/blob/master/apollo/now.json]. I've adapted it here for use with the official Apollo tutorial.
```
{
  "version": 2,
  "builds": [
    { "src": "src/*.js", "use": "@now/node-server" }
  ],
  "routes": [
    { "src": "/.*", "dest": "index.js" }
  ]
}
```

However, I'm still seeing issues with compiling modules using ncc.
```
compiling entrypoint with ncc...
11/25 08:18 PM (38m)
ncc: Module directory "/tmp/673d85ef/user/node_modules/subscriptions-transport-ws/node_modules/ws/lib" attempted to require "utf-8-validate" but could not be resolved, assuming external.
11/25 08:18 PM (38m)
ncc: Module directory "/tmp/673d85ef/user/node_modules/subscriptions-transport-ws/node_modules/ws/lib" attempted to require "bufferutil" but could not be resolved, assuming external.
11/25 08:18 PM (38m)
ncc: Module directory "/tmp/673d85ef/user/node_modules/sequelize/lib/dialects/postgres" attempted to require "pg-hstore" but could not be resolved, assuming external.
11/25 08:18 PM (38m)
ncc: Module directory "/tmp/673d85ef/user/node_modules/sequelize/lib/dialects/mysql" attempted to require "mysql2" but could not be resolved, assuming external.
11/25 08:18 PM (38m)
ncc: Module directory "/tmp/673d85ef/user/node_modules/sequelize/lib/dialects/mssql" attempted to require "tedious" but could not be resolved, assuming external.
11/25 08:18 PM (38m)
ncc: Module directory "/tmp/673d85ef/user/node_modules/sequelize/lib/dialects/mssql" attempted to require "tedious" but could not be resolved, assuming external.
11/25 08:18 PM (38m)
ncc: Module directory "/tmp/673d85ef/user/node_modules/sequelize/lib/dialects/postgres" attempted to require "pg" but could not be resolved, assuming external.
11/25 08:18 PM (38m)
[...]

PR #444, opened 2019-05-14
Remove "Deply to Now" from tutorial

joeysino:
This part of the tutorial is no longer working, as reported in https://github.com/apollographql/apollo/issues/264

As such, it only hampers someone trying to work through the tutorial.

Even following the suggestions in the issue, I was not able to create a successful deployment. (If I had, I would have opened a PR to fix the instructions.)

Since this seems to be producing a more negative experience than positive (and since it does not seem to be a critical part of the tutorial), I suggest removing this section until working instructions become available.


docs/source/tutorial/production.md
```diff
@@ -63,15 +63,3 @@ Publishing your schema to Apollo Engine unlocks many features necessary for runn
 We also want to be transparent that the features we just described, such as viewing specific execution traces and validating schema changes against recent operations, are only available on a paid plan. Individual developers just getting started with GraphQL probably don't need these features, but they become incredibly valuable as you're working on a team. Additionally, layering these paid features on top of our free developer tools like Apollo VSCode makes them more intelligent over time.
 
 We're committed to helping you succeed in building and running an Apollo graph API. This is why features such as publishing and downloading schemas from the registry, our open source offerings like Apollo Client and Apollo Server, and certain developer tools like Apollo VSCode and Apollo DevTools will always be free forever.
-
-## Deploy your graph API
-
-To deploy our app to Now, run the `now` command from the `server` directory of the app. The command may prompt you to login if you haven't already.
-
-```bash
-$ npx now
```
joeysino: According to [this comment](https://github.com/apollographql/apollo/issues/264#issuecomment-502449102), whilst `npx now` is suffering from problems, running `now` globally works fine.

So perhaps if we remove the `npx` prefix, this section can work.

(We might want to remind the reader that `npm install -g now` is needed, and let them know that `npx` may not work as expected.)

PR #446, opened 2019-05-17
Docs: Getting Started - Update to use now platform version 1

bassettsj:
Fixes incorrect instructions for now deployment since it will not work without `now.json` anymore


bassettsj:
Closing since this duplicates #264

PR #451, opened 2019-05-24
Fix broken GraphQL playground

subhojit777:
See https://github.com/apollographql/apollo-server/issues/2676


abernix:
Thanks for opening this, but we don't want to make this a permanent part of the documentation.  This was a temporary workaround to an error which should hopefully no longer be necessary.

PR #455, opened 2019-05-28
Rename operation-registry to operation-safelisting

daniman:
This is a spin-off of thoughts from #354.

We've been slowly transiting the Platform docs articles to describe workflows as opposed to features/fixtures; i.e. the "why" instead of the "what". Renaming this article to "Operation Safelisting" as opposed "Operation Registry" is in line with that plan, as the safelisting is the reason why one would opt into using the operation registry today.

Platform docs articles:
image

We may want to consider having multiple articles to represent the operation registry in the future, just like we have two to represent the schema registry ([Tracking your GraphQL schema](https://www.apollographql.com/docs/platform/schema-registry) and [Validating schema changes](https://www.apollographql.com/docs/platform/schema-validation)). We should do that when the time comes.

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -27,7 +27,7 @@ module.exports = {
             'platform/schema-registry',
             'platform/schema-validation',
             'platform/client-awareness',
-            'platform/operation-registry',
+            'platform/operation-safelisting',
             'platform/editor-plugins',
             'platform/performance',
             'platform/integrations'
--- a/docs/source/platform/operation-safelisting.md
+++ b/docs/source/platform/operation-safelisting.md
@@ -1,6 +1,6 @@
 ---
-title: Operation registry
-description: How to secure your graph with operation safelisting
+title: Operation safelisting
+description: How to secure your graph by enforcing a safelist of registered operations
 ---
 
 ## Overview
--- a/docs/static/_redirects
+++ b/docs/static/_redirects
@@ -1,7 +1,7 @@
 / /docs/
 
 # Redirect Removed Guides to Platform Features
-/docs/guides/security.html /docs/platform/operation-registry.html
+/docs/guides/security.html /docs/platform/operation-safelisting.html
 /docs/guides/monitoring.html /docs/platform/integrations.html
 /docs/guides/versioning.html /docs/platform/schema-validation.html#versioning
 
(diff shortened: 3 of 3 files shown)


abernix:
More discoverable in that it doesn't match the name of the plugin itself?

abernix:
Reverted in #463 because this would still need a redirect to be mergeable per our existing patterns.  Also, perhaps it would be better to have the side-bar say "Securing your graph", rather than either _Operation registry_ or _Operation safelisting_?  Perhaps we should update the image on the main docs page to not say _Operation registry_ at the same time (otherwise the documentation for that planet seems hard to find)?:

image

daniman:
Agree that "secure your graph" is a good name as well. That's actually what Peggy and I sketched out in a Platform docs nav outline back in December: https://apollographql.quip.com/cvpZAedYnvTo/Platform-docs-work. Going with "operation safelisting" seemed like a smaller change in that direction, but we could just go all the way.

There is a redirect in there, does it not get picked up the way it used to?

daniman:
This picture doesn't need to change because the operation registry, the schema registry, and the trace warehouse are the 3 core tenets of the platform. Safe-listing is just our first feature built on top of the registry.

abernix:
It wasn't working for me!

abernix:
Yeah, the redirect just doesn't function, which was what I'd acted on when reverting it. 
 Apologies for very incorrectly making it sound like this PR was lacking the redirect.  It clearly tried, even if it didn't work (which is the fault of the theme.)

I'm going to open a follow-up PR proposal.

abernix:
@daniman 	

> This picture doesn't need to change because the operation registry, the schema registry, and the trace warehouse are the 3 core tenets of the platform. Safe-listing is just our first feature built on top of the registry.

Ok, I like that and it makes sense.  I do think we should try to make some more connections between those tenants and actual links on the sidebar.

image

For example, to me, I would hope to find some way to explore those tenants in a 1:1 manner from the "planets I see in the universe" to the links in the _Platform_ section of the sidebar.  Operation registry _had_ that connection before, and it suddenly did not.

The initial wholesale rename of the page seemed like it was going more in the direction of a product rename, which seemed premature and like it would necessitate other work which I didn't see happening.

I opened #465 which changes the title to a more workflow-based name than, from my perspective, even "Operation safe-listing" would have provided and aligns more with the descriptions of workflows on other platform features (i.e. "Schema registry" = _Tracking your GraphQL schema_ 👍 ).

daniman:
@abernix Each of the systems is described in that way in the Platform intro doc, which is the next page after the one that picture is on: https://www.apollographql.com/docs/intro/platform/#cloud-services. We could pull them out into their own articles in the future, but we don't have that for the schema registry and trace warehouse today either, so there isn't a natural place for that to fit that I see.

PR #456, opened 2019-05-29
Correct minimum node and npm versions

subhojit777:
The tutorial uses `async`/`await` syntax which was introduced in Node 7.6.0


abernix:
Thank you for this! I think we can go with the minimum "Long-Term Support" versions of Node.js and npm now that Node.js 6 (which shipped with npm 5 in its later days) is officially retired.  Node.js 8 ships with npm 6 in most recent versions, and to be honest, we don't test on npm v5, so I'd rather not steer anyone wrong.  Thank you again!

PR #463, opened 2019-05-31
Revert "Rename operation-registry to operation-safelisting"

abernix:
Reverts apollographql/apollo#455

As has always been our policy for moving documentation around, this should have a redirect setup for it to preserve existing links.  I'm reverting it because that should be done as part of the rename and currently, existing links would be broken because of this.  

Additionally, I do think we should discuss it first.


abernix:
Happy to consider this, but the redirects should be in place.

PR #466, opened 2019-05-31
Document how to configure schemaTag in operation registry plugin - AP-468

cheapsteak:
(Note: can merge after https://github.com/apollographql/apollo-tooling/pull/1307 is released. update: released)

This PR adds 2 optional sub-steps in the setup flow to demonstrate how to configure the schema tag for the operation registry plugin and the cli  

It also removes the "Only the default schema tag (`current`) is supported." section from the "Limitations" section

### DIFF
--- a/docs/source/platform/operation-registry.md
+++ b/docs/source/platform/operation-registry.md
@@ -118,6 +118,10 @@ Currently, once an operation is registered it will remain registered indefinitel
 
 If you encounter any errors, check the _**Troubleshooting**_ section below.
 
+####3.1 Optionally, set the schema tag  
+
+To specify the schema tag to register operations on, pass an additional `--tag <TAG>` argument (`npx apollo client:push --tag <TAG>`).
+
 ### 4. Disable subscription support on Apollo Server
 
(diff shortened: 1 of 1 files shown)


docs/source/platform/operation-registry.md
```diff
@@ -168,6 +165,23 @@ const server = new ApolloServer({
...
+```js
+const server = new ApolloServer({
+  // Existing configuration
+  plugins: [
+    require("apollo-server-plugin-operation-registry")({
+      schemaTag: 'overrideTag' // highlight-line
+    })
+  ]
+});
+```
```
evans: We should also mention the need to set the tag in the client push command here in order to get the operations into the manifest the plugin pulls

docs/source/platform/operation-registry.md
```diff
@@ -168,6 +169,23 @@ const server = new ApolloServer({
 });
 ```
 
+####5.1 Optionally, set the schema tag
+
+The plugin will automatically pickup the value of the environment variable `ENGINE_SCHEMA_TAG` and use that as the schema tag.
```
evans: If users specify this env var, then our new changes will be breaking, since they'll pick up a new manifest. Let's make sure this is in the upgrade guide

cheapsteak: I will remove this for now to match changes in https://github.com/apollographql/apollo-platform-commercial/pull/125

Thanks for picking up on this 🙏

PR #469, opened 2019-06-01
Spell checker PoC

trevor-scheer:
I know we've only discussed this in passing, but I was interested in taking a crack at it. If either of you (@peggyrayzis, @trevorblades) have a different direction in mind for solving this problem, please don't hesitate to shut down this PR 👍 it was more of a curiosity and exploration than anything else.

This PR implements a spell checking tool against `.md` and (mostly) `.mdx` files.
The caveat with `.mdx` is that I think the checker finds some false positives like `_You` (maybe due to formatting?). Not a huge deal.

To run: `npm run spell-check`
The tool itself is easy to use. It provides an interactive mode for correcting mistakes and adding / ignoring words to the dictionary.

I was hoping for some other packages to try out, but didn't find anything interesting. There's one major thing missing with this tool, and that's ignoring by regex. A lot of things are ignored repeatedly in the `.spelling` file, i.e.:
`index.js`, `resolvers.js`, `launch.js` ...
`v1`, `v1.0`, `v2.x` ...

While this is an annoyance, it's more of a one-time, small cost to incur (and it's already done in this PR). A next / final step for this PR is to run the linter in CI (using the `--report` flag)
`npm run spell-check:ci`

This is already set-up (but untested, as I don't want to set up CircleCI on this repo to test it unless we decide we want to go forward with this).

The API for reference:
https://github.com/lukeapage/node-markdown-spellcheck#readme


trevor-scheer:
Update after conversation with @trevorblades - we're going to explore a Gatsby remark plugin to solve this.

PR #470, opened 2019-06-02
Delete references to Zeit Now

mrenrich84:
Some ref were still left to that step


StephenBarlow:
Thanks for the PR @mrenrich84! The reference in question has been removed.

PR #474, opened 2019-06-07
Op-reg / safelist migration guide

cheapsteak:
image
image

### DIFF
--- a/docs/source/platform/operation-registry.md
+++ b/docs/source/platform/operation-registry.md
@@ -290,6 +290,35 @@ const server = new ApolloServer({
 });
 ```
 
+## Migrating from 0.1
+
+#### On servers that safelist operations
+
+- Upgrade `apollo-server` to `2.6.2` and `apollo-server-plugin-operation-registry` to `0.2.0-alpha.1`
+
+#### To target graph variants other than "current"
+
+- Upgrade the `apollo` CLI package to 2.13.0
+
+- Ensure that a schema has been published to the specified graph variant.
+  This can be done by running `apollo service:push --tag <TAG>`
+
+- Ensure that operations have been registered to the specified graph variant.
+  This can be done by running `apollo client:push --tag <TAG>`
+
+- Set the `schemaTag` field of `apollo-server-plugin-operation-registry` to the targeted graph variant
+
+```js
+const server = new ApolloServer({
+  // Existing configuration
+  plugins: [
+    require("apollo-server-plugin-operation-registry")({
+      schemaTag: 'prod' // highlight-line
+    })
+  ]
+});
+```
+
 ## Troubleshooting
 
 #### The server indicates `Access denied.` (or `AccessDenied`) when fetching the manifest
--- a/docs/source/platform/operation-registry.md
+++ b/docs/source/platform/operation-registry.md
@@ -234,7 +234,7 @@ Finally, to confirm that the server will allow permitted operations, try running
 
 In some cases, deployments may want to selectively enable the behavior of `forbidUnregisteredOperations` depending on environmental conditions (e.g. based on headers).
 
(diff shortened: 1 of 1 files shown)


docs/source/platform/operation-registry.md
```diff
@@ -290,6 +290,50 @@ const server = new ApolloServer({
 });
 ```
 
+## Migrating from 0.1
+
+#### On servers that safelist operations
+
+- Upgrade `apollo-server` to `2.6.2` and `apollo-server-plugin-operation-registry` to `0.2.0-alpha.1`
+
+#### To target graph variants other than "current"
```
evans: I think it would be helpful to add that "current" is the default, no tag. Maybe something like:

```
#### To target graph variants other than "current"(default, no tag)
```

docs/source/platform/operation-registry.md
```diff
@@ -290,6 +290,50 @@ const server = new ApolloServer({
 });
 ```
 
+## Migrating from 0.1
```
evans: Since we're going from alpha to alpha, i think it would be more clear to offer 0.1-alpha.4

docs/source/platform/operation-registry.md
```diff
@@ -290,6 +290,50 @@ const server = new ApolloServer({
...
+
+#### On servers that safelist operations
+
+- Upgrade `apollo-server` to `2.6.2` and `apollo-server-plugin-operation-registry` to `0.2.0-alpha.1`
```
evans: We should mention that there are no breaking changes when using, since I think that is a large concern

docs/source/platform/operation-registry.md
```diff
@@ -290,6 +290,50 @@ const server = new ApolloServer({
...
+  // Existing configuration
+  plugins: [
+    require("apollo-server-plugin-operation-registry")({
+      schemaTag: 'prod' // highlight-line
+    })
+  ]
+});
+```
+
+- It might be advisable to do an initial test run with `debug` and `dryRun` both on
```
evans: I think we can be stronger here, since we're doing this internally. Maybe: "An initial test run with `debug` and `dryRun` enabled is suggested"

docs/source/platform/operation-registry.md
```diff
@@ -234,7 +234,7 @@ Finally, to confirm that the server will allow permitted operations, try running
 
 In some cases, deployments may want to selectively enable the behavior of `forbidUnregisteredOperations` depending on environmental conditions (e.g. based on headers).
 
-To selectively enable operation safelisting, the `forbidUnregisteredOperations` setting supports a [predicate function](https://en.wikipedia.org/wiki/Predicate_(mathematical_logic) which receives the request context and can return `true` or `false` to indicate whether enforcement is enabled or disabled respectively.
+To selectively enable operation safelisting, the `forbidUnregisteredOperations` setting supports a [predicate function](https://en.wikipedia.org/wiki/Predicate_(mathematical_logic)) which receives the request context and can return `true` or `false` to indicate whether enforcement is enabled or disabled respectively.
```
evans: 😍

docs/source/platform/operation-registry.md
```diff
@@ -290,6 +290,50 @@ const server = new ApolloServer({
 });
 ```
 
+## Migrating from 0.1-alpha.4
+
+#### On servers that safelist operations
+
+- Upgrade `apollo-server` to `2.6.2` and `apollo-server-plugin-operation-registry` to `0.2.0-alpha.1` (there are no breaking changes)
```
evans: When I first read "there are no breaking changes", I struggled to connect with how it was relevant. I think it is hanging without an action/attitude that we want users to adopt(confidence in the change/assuring that no additional updates are needed). I recommend adding a statement about why it's relevant. In this case, something like "so the upgrade can be made without additional work"

docs/source/platform/operation-registry.md
```diff
@@ -290,6 +290,50 @@ const server = new ApolloServer({
...
+
+#### To target graph variants other than "current" (default, no tag)
```
evans: Does it make sense to connect "current" and (default, no tag) more? Maybe with no space between?

Totally could be nitpicky, since it's a small optimization. Also might not be possible to misconstrue what in the title is the default/no tag, so feel free to ignore

PR #476, opened 2019-06-08
Removed incorrect schema block

ntziolis:
This section describes how to create a resolver for a field on the mission type. However the removed block references creating a new mutation that was never mentioned previously or after.


abernix:
Resolved by #494.

PR #480, opened 2019-06-12
op-reg: update migration guide with more details

evans:
In order to fully describe what the upgrade contains and requires,
we need to expand on the operation registry migration plan to include a
feature description and more detailed instruction set.


docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+
+      // suggested before enforcing the safelist
+      debug: true,
+      dryRun: true,
+    })
+  ]
+});
+```
+
+> Note: in order to copy the operations from one tag to another, please contact the Apollo team through Intercom or your shared Slack channel
```
daniman: I think you can say something more direct here, like "If you want to copy all previously registered operations to a new tag, please reach out to the Apollo team". I would remove the part of this message that describes the method, since not everyone has a shared Slack channel with us.

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+operation registration and schema uploads, since a manifest needed to be
+created for each schema publish with a new hash. Concretely the GCS url was:
+
+`/(hash of service id)/(schema hash)/manifest.json`
+
+The new model removes this coupling by using the Apollo api key as the shared
+secret. In order to enable independent secret rotation, the api key is used
+to reference a storage secret that then references the manifests
+
+`/(service id)/(hash of api key)/storage-secret.json` <br>
```
cheapsteak: Not sure if this matters, but we've changed this to `/(service id)/storage-secret/(hash of api key).json`

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+These changes should be transparent, since the new version of the operation
+registry plugin, `0.2.0-alpha.1`, will attempt to use the new scheme and then
+fallback on the old. All apollo cli versions write to the new
+and old manifest locations without additional configuration.
+
+#### Operation registration observability
+
+To improve the client side registration, `apollo client:push` now includes
+the file name and location for operations added and connects the error
+message with the operation that fails.
```
daniman: What's does the experience look like? Can we add a sample output maybe, for those of us just reading and not testing the new commands directly?

docs/source/platform/operation-registry.md
evans: Good catch! Thank you

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+### Upgrade path
+
+While this upgrade should not require additional work, we recommend running
+the new plugin with `dryRun` and `debug` enabled. To report metrics and use
+the new storage location, upgrade:
+
+- `apollo-server` or `apollo-server-<variant>` to `2.6.3`
+- `apollo-server-plugin-operation-registry` to `0.2.0-alpha.1`.
+
+If operations have not been registered since June 6th, the plugin will
```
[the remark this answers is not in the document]
evans: Good thought!

docs/source/platform/operation-registry.md
evans: Love it! I added some copy around the pictures to have them in context 🎉 
 
image

docs/source/platform/operation-registry.md
```diff
@@ -118,7 +118,7 @@ Currently, once an operation is registered it will remain registered indefinitel
 
 If you encounter any errors, check the _**Troubleshooting**_ section below.
 
-####3.1 Optionally, set the schema tag
```
zionts: It might be nice to elucidate the **why** or link to another article, rather than only the **how**

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+
+While this upgrade should not require additional work, we recommend running
+the new plugin with `dryRun` and `debug` enabled. To report metrics and use
+the new storage location, upgrade:
+
+- `apollo-server` or `apollo-server-<variant>` to `2.6.3`
+- `apollo-server-plugin-operation-registry` to `0.2.0-alpha.1`.
+
+If operations have not been registered since June 6th, the plugin will
+fallback on the old location. With `debug` enabled, the plugin will log which
```
zionts: tangential: Is there a way to pass a custom logger?

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+manifest is used and the operations added.
+
+#### To target variants/tags other than "current"(default, no tag)
+
+Operations are now registered against a variant/tag and retrieved from the
```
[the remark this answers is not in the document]
zionts: Agreed. I love that we offer a one-time migration to users that reach out to us -- maybe we can say that next to the strong warning to both highlight the severity and show our love

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+manifest specific to a variant/tag. When a variant/tag is set, the operation
```
zionts: I think this paragraph should have the converse (i.e. "when a variant/tag is left unspecified, the tools will register and pull using the default tag, "current")

zionts: Also, wording nit: I'm not sure how useful it is to oscillate back and forth with the wording of tag / variant here. I think it might be easier just to exclusively use "tag" for now since that's what it is in all the config.

docs/source/platform/operation-registry.md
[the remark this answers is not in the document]
evans: True! I also describe what a tag/variant is, since it might not be totally clear

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+The new model removes this coupling by using the Apollo api key as the shared
+secret. In order to enable independent secret rotation, the api key is used
+to reference a storage secret that then references the manifests
+
+`/(service id)/(hash of api key)/storage-secret.json` <br>
+`/(service id)/(storage secret)/(tag)/manifest.v2.json`
+
+These changes should be transparent, since the new version of the operation
+registry plugin, `0.2.0-alpha.1`, will attempt to use the new scheme and then
+fallback on the old. All apollo cli versions write to the new
```
zionts: this fallback has pretty real implications as you migrate to tag usage. It means that if you've never registered an operation to the `prod` tag, then things will work fine at first, but then once you register one operation, that will be the only one allowed. I'm not sure how to communicate the nuance of the fallback strategy here, but I just wanted to bring it up.

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+and old manifest locations without additional configuration.
```
zionts: I think we might want to specify that tag-specific manifests are only created once an operation is pushed to a tag.

docs/source/platform/operation-registry.md
```diff
@@ -386,3 +342,126 @@ By clicking on the URL listed in the `Checking for manifest changes at` message,
...
+
+##### Metrics page
+
+The filter on the metrics page will provide the option to show unregistered operations only
+
+<img src="../img/operation-registry/metrics-filter.png" width="50%" style="margin: 5%" alt="The metrics filter showing the option to select unregistered operations">
+
+#### Variant/tag awareness
+
+Operations are now registered against a variant/tag and retrieved from the
```
zionts: Another opportunity to have a warning message (in case people skip around)

docs/source/platform/operation-registry.md
```diff
@@ -365,18 +365,24 @@ the new storage location, upgrade:
...
-manifest specific to a variant/tag. When a variant/tag is set, the operation
+manifest specific to a variant/tag. This does not change the default behavior
+of the Apollo platform, which universally uses the variant/tag `current` when
+unspecified. Variants/tags represent different environments, such as
+`staging`, `test` or `prod`, so they do not share data. This means that
+operations must be re-registered when using or transitioning to a new
+variant/tag.
+
+When a variant/tag is set in `apollo client:push`, the operation
 will be validated against the latest schema published under that variant/tag
```
zionts: I thought we took out operation validation, no?

docs/source/platform/operation-registry.md
```diff
@@ -389,17 +395,17 @@ To register operations under a specific variant/tag:
...
       debug: true,
-      dryRun: true,
+      dryRun: true
     })
   ]
 });
 ```
 
-> Note: in order to copy the operations from one tag to another, please contact the Apollo team through Intercom or your shared Slack channel
+> Note: If you want to copy all previously registered operations to a new variant/tag, please contact the Apollo team
```
zionts: nit: I think this could be a little friendlier of an offer for help. Something like
> We know it can be challenging to communicate with all consumers of the graph to publish their operations to a new tag. Reach out to the Apollo team if you want to move to a tag-based registry, and we'll be happy to migrate all operations registered to the `current` tag to the tag of your choosing to make that process easier.

zionts:
I think this looks great @evans! I think the main bit of 🤷‍♂ from me is that it feels a bit weird to be using the term `tag/variant` in our copy when all of the configuration and images are still `tag`. I totally understand why (and I try to do the same internally), but I think it might cause more external confusion than help on-board our users onto the new terminology. Maybe it's good for future-proofing docs, though? I'm not sure.

docs/source/platform/operation-registry.md
evans: We [validate the operation](https://github.com/mdg-private/monorepo/blob/5ed1482a25ba0a57dbc50fab0197c5cea119de4b/mdg/engine/api/gateway/src/operationRegistryService/index.ts#L301-L304) and fail the publish if they don't validate

evans:
> I think this looks great @evans! I think the main bit of 🤷‍♂ from me is that it feels a bit weird to be using the term `tag/variant` in our copy when all of the configuration and images are still `tag`. I totally understand why (and I try to do the same internally), but I think it might cause more external confusion than help on-board our users onto the new terminology. Maybe it's good for future-proofing docs, though? I'm not sure.

Right now we use them interchangeably as I understand it, so I'm placing them together to hopefully reduce confusion. On [most? of the UI](https://meteor.slack.com/archives/C03EVP694/p1560357532017300) and rest of the docs, we use variant. The cli as I understand it is the last place we use tag

docs/source/platform/operation-registry.md
evans: ❤️ me gusta!

docs/source/platform/operation-registry.md
evans: not yet! Though that would be an interesting feature

docs/source/platform/operation-registry.md
evans: good point! I'll add some more language around the fallback being only the operations for the current tag

PR #485, opened 2019-06-17
Change missionPatch type to String

Pringels:
`missionPatch`'s type in the schema must have accidentally been changed from `String` to `PatchSize` when a resolver was added.

This causes a fairly difficult to debug error in the Client UI while working through the tutorial. (at least for graphQL newbies like me 😛 )


evans:
@Pringels Thank you so much!

PR #488, opened 2019-06-20
Wrong text case on resolvers code.

douglasmoraisdev:
dataSources.launchAPI ans dataSources.userAPI writed with wrong text case on resolvers code.

### DIFF
--- a/docs/source/tutorial/resolvers.md
+++ b/docs/source/tutorial/resolvers.md
@@ -70,10 +70,10 @@ _src/resolvers.js_
 module.exports = {
   Query: {
     launches: (_, __, { dataSources }) =>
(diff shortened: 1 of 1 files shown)


evans:
This looks okay to me, since the [data sources are defined as lower case](https://github.com/apollographql/apollo/blame/f310f73a714fd2c146cf58511a0674aa775229ab/docs/source/tutorial/resolvers.md#L49) and the [final output of the tutorial matches](https://github.com/apollographql/fullstack-tutorial/blob/d780ec0ceed274fdc296eebfaf20d54499e8ea31/final/server/src/resolvers.js#L28). Thanks for opening this! I'm going to close it for now, though feel free to open another with some more description if you think the change should still be made

PR #503, opened 2019-07-02
[WIP] Federation: Stub PR to update `service:check`

zionts:
This PR should update the `schema-validation` docs to include
information and instructions on the feature for federation. This update
should include the added model and note that validation can be used to
validate service **addition** as well (i.e. you can validate that a new
service will compose with the existing without affecting the graph)

### DIFF
--- a/docs/source/platform/schema-validation.mdx
+++ b/docs/source/platform/schema-validation.mdx
@@ -191,7 +191,7 @@ Since breaking changes are detected using live traffic, your service will _need 
 
 ## Set up schema validation
 
-To set up schema validation, you wlil need to be both actively sending traces and registering schemas to Apollo:
+To set up schema validation, you will need to be both actively sending traces and registering schemas to Apollo:
 
 1. [Set up trace reporting to Apollo Engine](/references/setup-analytics/)
 1. [Set up schema registration in your continuous delivery pipeline](/platform/schema-registry/)
--- a/docs/source/platform/schema-validation.mdx
+++ b/docs/source/platform/schema-validation.mdx
@@ -7,7 +7,7 @@ description: Check if schema changes are safe or breaking by comparing against l
 [//]: # (Assignee: Caydie)
 [//]: # (Reviewer: Jake)
 
-There are many types of schema changes that can be potentially breaking to clients, like removing a field, if made without special consideration. For safety, some organizations take the approach of _never_ making these types of changes, but this leads to an ever-growing schema and reduced API flexibility over time. In reality, making these types of changes to a schema can be very safe as long as you have tools in place to ensure that no queries are broken in the process.
+There are many types of schema changes that can be potentially breaking to clients, like removing a field, if made without special consideration. For safety, some organizations take the approach of _never_ making these types of changes, but this leads to an ever-growing schema and reduced API flexibility over time. In the case of federated services where multiple organizations or teams are working on different parts of the same graph, these kinds of breaking changes are especially critical to monitor. In reality, making these types of changes to a schema can be very safe as long as you have tools in place to ensure that no queries are broken in the process.
 
 Apollo provides a tool to protect for exactly this scenario called **schema validation**.
 
@@ -19,7 +19,7 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
 ## How it works
 
 Schema validation is run through the Apollo CLI by executing the `apollo service:check` command. Apollo will generate a diff between your local schema and your most recently registered schema, then validate that the changes are safe by checking if any queries actively running against your graph will be affected.
-If your graph is composed using federation, you can also use the `apollo service:check` command to validate individual service changes by providing a service name.
+If your graph is composed using federation, you can also use the `apollo service:check` command to validate individual service changes by providing a service name. However, unlike running service check on a non-federated service, Engine first needs to run composition on the proposed partial schema with the current list of implementing services and their partial schemas. Composition must pass before validation can continue.
 
 Here's how it works:
 
@@ -32,9 +32,22 @@ Here's how it works:
 1. Engine returns the schema diff and indicates any breaking changes found.
 1. The CLI prints the output of this check with a link to view more details in the Engine UI.
 
+Here's how it works for federation:
+
+1. You _also_ run `apollo service:check` but instead you provide a service name like so: `apollo service:check --federated --serviceName=example --endpoint=http://localhost:4000/graphql`
+ The proposed partial schema for that provided service is then sent to Engine's schema registry where it will attempt to compose the partial schema with the current schema.
+1. If composition fails then we do not continue with validation and terminate early.
+1. If composition succeeds Engine then creates a diff between the local partial schema and the most recently published schema in the registry.
+1. Engine walks through the partial schema diff change-by-change and compares against the operation list see if the changes will affect the behavior of any operations.
+1. Engine returns the schema diff and indicates any breaking changes found.
+1. The CLI prints the output of this check with a link to view more details in the Engine UI. 
+
 ### Breaking change detection
 
 // [TODO: Add a federation-specific paragraph about composition errors]
+
+Since federation involves merging the partial schema you're running service check against with the current schema, it's possible to run into compposition errors. For example, if you try to run validation on a partial schema with conflicting type names as the current schema then this will definitely result in a composition error. 
+At the moment composition warnings are also grouped together with composition errors. Composition warnings might involve referencing another service's deprecated field.
 
 Not all schema changes are potentially breaking. Some changes, like adding a field, will always be safe and never cause unexpected behavior for active clients. Other changes, like removing a field or changing a return type, can potentially affect the behavior of clients making queries that use those fields. These are what we consider potentially breaking changes.
 
--- a/docs/source/platform/schema-validation.mdx
+++ b/docs/source/platform/schema-validation.mdx
@@ -32,8 +32,6 @@ Here's how schema validation works for federated and non-federated services:
 
 ### Breaking change detection
 
-// [TODO: Add a federation-specific paragraph about composition errors]
-
 Since federation involves merging the partial schema you're running service check against with the current schema, it's possible to run into compposition errors. For example, if you try to run validation on a partial schema with conflicting type names as the current schema then this will definitely result in a composition error. 
 
 Not all schema changes are potentially breaking. Some changes, like adding a field, will always be safe and never cause unexpected behavior for active clients. Other changes, like removing a field or changing a return type, can potentially affect the behavior of clients making queries that use those fields. These are what we consider potentially breaking changes.
--- a/docs/source/platform/schema-validation.mdx
+++ b/docs/source/platform/schema-validation.mdx
@@ -7,7 +7,7 @@ description: Check if schema changes are safe or breaking by comparing against l
 [//]: # (Assignee: Caydie)
 [//]: # (Reviewer: Jake)
 
-There are many types of schema changes that can be potentially breaking to clients, like removing a field, if made without special consideration. For safety, some organizations take the approach of _never_ making these types of changes, but this leads to an ever-growing schema and reduced API flexibility over time. In the case of federated services where multiple organizations or teams are working on different parts of the same graph, these kinds of breaking changes are especially critical to monitor. In reality, making these types of changes to a schema can be very safe as long as you have tools in place to ensure that no queries are broken in the process.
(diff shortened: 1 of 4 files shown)


docs/source/platform/schema-validation.mdx
```diff
@@ -3,27 +3,52 @@ title: Validating schema changes
...
 
+
 ## How it works
 
 Schema validation is run through the Apollo CLI by executing the `apollo service:check` command. Apollo will generate a diff between your local schema and your most recently registered schema, then validate that the changes are safe by checking if any queries actively running against your graph will be affected.
+If your graph is composed using federation, you can also use the `apollo service:check` command to validate individual service changes by providing a service name. However, unlike running service check on a non-federated service, Engine first needs to run composition on the proposed partial schema with the current list of implementing services and their partial schemas. Composition must pass before validation can continue.
 
 Here's how it works:
 
+// [TODO: Hopefully make this a toggle, where you can see the federation and the non-federation version of how it works].
```
zionts: Let's forgo the toggle idea for now since adding in interactivity should align with our design paradigms that we're still trying to nail down.

docs/source/platform/schema-validation.mdx
```diff
@@ -3,27 +3,52 @@ title: Validating schema changes
...
 1. You run `apollo service:check` locally or in CI. The proposed schema is sent to Engine's schema registry.
 1. Engine creates a diff between the local schema and the most recently published schema in the registry.
 1. Engine fetches a list of all operations sent to your graph in the last day (time window is [configurable](#adjusting-validation-parameters)).
 1. Engine walks through the schema diff change-by-change and compares against the operation list to see if the changes will affect the behavior of any operations.
 1. Engine returns the schema diff and indicates any breaking changes found.
 1. The CLI prints the output of this check with a link to view more details in the Engine UI.
 
+Here's how it works for federation:
+
+1. You _also_ run `apollo service:check` but instead you provide a service name like so: `apollo service:check --federated --serviceName=example --endpoint=http://localhost:4000/graphql`
```
zionts: You don't need to have `--federated` anymore. I also think that we should just have an additional **extra** step in the above flow rather than repeating all the same flow here.

docs/source/platform/schema-validation.mdx
```diff
@@ -3,27 +3,52 @@ title: Validating schema changes
...
+1. Engine walks through the partial schema diff change-by-change and compares against the operation list see if the changes will affect the behavior of any operations.
+1. Engine returns the schema diff and indicates any breaking changes found.
+1. The CLI prints the output of this check with a link to view more details in the Engine UI. 
+
 ### Breaking change detection
 
+// [TODO: Add a federation-specific paragraph about composition errors]
+
+Since federation involves merging the partial schema you're running service check against with the current schema, it's possible to run into compposition errors. For example, if you try to run validation on a partial schema with conflicting type names as the current schema then this will definitely result in a composition error. 
+At the moment composition warnings are also grouped together with composition errors. Composition warnings might involve referencing another service's deprecated field.
```
zionts: there are no more composition warnings.

peggyrayzis:
Just talked this over with @zionts. For all federation content, can we please make a separate heading called "Val

docs/source/platform/schema-validation.mdx
caydie-tran: oh man that's a v good idea :)

docs/source/platform/schema-validation.mdx
caydie-tran: deleto

docs/source/platform/schema-validation.mdx
caydie-tran: okie dokie :)

docs/source/platform/schema-validation.mdx
```diff
@@ -12,18 +16,24 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
 ## How it works
 
 Schema validation is run through the Apollo CLI by executing the `apollo service:check` command. Apollo will generate a diff between your local schema and your most recently registered schema, then validate that the changes are safe by checking if any queries actively running against your graph will be affected.
+If your graph is composed using federation, you can also use the `apollo service:check` command to validate individual service changes by providing a service name. However, unlike running service check on a non-federated service, Engine first needs to run composition on the proposed partial schema with the current list of implementing services and their partial schemas. Composition must pass before validation can continue.
```
JakeDawkins: > If your graph is composed using federation, you can also use the `apollo service:check` command to validate individual service changes by providing a service name.

I'd add that the service name is a flag, and but it in backticks. Like...

> If your graph is composed using federation, you can also use the `apollo service:check` command to validate individual service changes by adding the `--serviceName` flag.

caydie-tran: great idea! :) added ~

docs/source/platform/schema-validation.mdx
```diff
@@ -3,7 +3,11 @@ title: Validating schema changes
...
+[//]: # (Assignee: Caydie)
+[//]: # (Reviewer: Jake)
+
+There are many types of schema changes that can be potentially breaking to clients, like removing a field, if made without special consideration. For safety, some organizations take the approach of _never_ making these types of changes, but this leads to an ever-growing schema and reduced API flexibility over time. In the case of federated services where multiple organizations or teams are working on different parts of the same graph, these kinds of breaking changes are especially critical to monitor. In reality, making these types of changes to a schema can be very safe as long as you have tools in place to ensure that no queries are broken in the process.
```
justinanastos:
```suggestion
There are many types of schema changes that can be potentially breaking to clients. For safety, some organizations take the approach of _never_ making these types of changes, but this leads to an ever-growing schema and reduced API flexibility over time. In the case of federated services where multiple organizations or teams are working on different parts of the same graph, these kinds of breaking changes are especially critical. Making these types of changes can be safe as long as you have tools in place.
```

This is removing a lot of the words to hopefully say the same thing

docs/source/platform/schema-validation.mdx
```diff
@@ -12,18 +16,24 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
 ## How it works
 
 Schema validation is run through the Apollo CLI by executing the `apollo service:check` command. Apollo will generate a diff between your local schema and your most recently registered schema, then validate that the changes are safe by checking if any queries actively running against your graph will be affected.
+If your graph is composed using federation, you can also use the `apollo service:check` command to validate individual service changes by adding the `--serviceName` flag. However, unlike running service check on a non-federated service, Engine first needs to run composition on the proposed partial schema with the current list of implementing services and their partial schemas. Composition must pass before validation can continue.
```
justinanastos:
```suggestion
With a federated graph, you will use the `apollo service:check` command to validate individual service changes by adding the `--serviceName` flag. Engine will run composition on the proposed partial schema with the current list of implementing services and their partial schemas, making sure that the composition is successful. That composed schema will then be diff'ed against the most recently registered schema and validate that those changes are safe.
```

docs/source/platform/schema-validation.mdx
```diff
@@ -12,18 +16,24 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
...
+
+But wait... What exactly is composition? 🤔 In order to have different implementing services contribute to the same graph, there needs to be a way to combine each implementing service's partial schemas into one large data graph. Engine's composition algorithm does exactly this.
```
justinanastos:
```suggestion
What exactly is composition? There needs to be a way to combine each implementing service's partial schema into one data graph; that's what Engine's composition algorithm does.
```

docs/source/platform/schema-validation.mdx
```diff
@@ -12,18 +16,24 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
...
 
-Here's how it works:
+Here's how schema validation works for federated and non-federated services:
 
-1. You run `apollo service:check` locally or in CI. The proposed schema is sent to Engine's schema registry.
-1. Engine creates a diff between the local schema and the most recently published schema in the registry.
+1. You run `apollo service:check` locally or in CI. For federated services run the same command but add `--serviceName=ExampleService --endpoint=http://loocalhost:4000/graphql` to the end. The proposed schema/partial schema (if using federation) is sent to Engine's schema registry.
```
justinanastos:
```suggestion
1. Run `apollo service:check` locally or in CI. For federated services, add `--serviceName=ExampleService --endpoint=http://loocalhost:4000/graphql`. The proposed schema, or partial schema if using federation, is then sent to Engine's schema registry.
```

docs/source/platform/schema-validation.mdx
```diff
@@ -12,18 +16,24 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
...
+1. If you're using federation, Engine's schema registry will attempt to compose the partial schema with the current schema. If composition fails then we do not continue with validation and exit early.
```
justinanastos:
```suggestion
1. If using federation, Engine's schema registry will attempt to compose the partial schema with the current schema. If composition fails, then we do not continue with validation.
```

docs/source/platform/schema-validation.mdx
```diff
@@ -12,18 +16,24 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
...
+1. Engine creates a diff between the local schema and the most recently published schema in the registry. For federation this step happens after composition passes.
```
justinanastos:
```suggestion
1. Engine creates a diff between the local schema and the most recently published schema in the registry.
```

Do we need the last sentence?

docs/source/platform/schema-validation.mdx
```diff
@@ -12,18 +16,24 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
...
 1. Engine fetches a list of all operations sent to your graph in the last day (time window is [configurable](#adjusting-validation-parameters)).
 1. Engine walks through the schema diff change-by-change and compares against the operation list to see if the changes will affect the behavior of any operations.
 1. Engine returns the schema diff and indicates any breaking changes found.
 1. The CLI prints the output of this check with a link to view more details in the Engine UI.
 
 ### Breaking change detection
 
+Since federation involves merging the partial schema you're running service check against with the current schema, it's possible to run into compposition errors. For example, if you try to run validation on a partial schema with conflicting type names as the current schema then this will definitely result in a composition error.
```
justinanastos:
```suggestion
The first breaking change Engine can capture is a federated service being unable to compose with the rest of a federated graph. Types of composition failures are documented [here]().
```

I feel like there's a lot of "because of this" that we don't need. I think this calls for more of the _what_ than the _why_.

- [ ] This needs the link to the composition failure docs

docs/source/platform/schema-validation.mdx
caydie-tran: less is more :) i like it

docs/source/platform/schema-validation.mdx
caydie-tran: meh no we don't need it. i think it's clear enough from the step above

docs/source/platform/schema-validation.mdx
caydie-tran: where are the composition failures doc?

docs/source/platform/schema-validation.mdx
```diff
@@ -12,10 +16,14 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
 ## How it works
 
 Schema validation is run through the Apollo CLI by executing the `apollo service:check` command. Apollo will generate a diff between your local schema and your most recently registered schema, then validate that the changes are safe by checking if any queries actively running against your graph will be affected.
+With a federated graph, you will use the `apollo service:check` command to validate individual service changes by adding the `--serviceName` flag. Engine will run composition on the proposed partial schema with the current list of implementing services and their partial schemas, making sure that the composition is successful. That composed schema will then be diff'ed against the most recently registered schema and validate that those changes are safe.
```
zionts: Should we have this as a separate paragraph? Or if it already is, maybe we should add a newline here for style nit.

zionts: It seems like the rest of this document is written in third person. Could we change this to be the same?

docs/source/platform/schema-validation.mdx
```diff
@@ -12,10 +16,14 @@ Apollo provides a tool to protect for exactly this scenario called **schema vali
...
+
+What exactly is composition? Composition is just Engine's algorithm to combine each implementing service's partial schema into one data graph.
```
zionts: I'm not sure that we need this paragraph here? Maybe we can link out from the `composition` keyword to the docs or the open source implementation? Also it's not Engine's algorithm because it's open source.

caydie-tran: i'll remove!

PR #505, opened 2019-07-02
Federation: Update operation registry docs [AP-587]

zionts:
Nothing here should change much, but we should verify that users don't
need to do anything to use the operation registry, and if they do, then
we should document those.

### DIFF
--- a/docs/source/platform/operation-registry.mdx
+++ b/docs/source/platform/operation-registry.mdx
@@ -3,6 +3,12 @@ title: Securing your graph
 description: How to secure your graph by enforcing a safelist of registered operations
 ---
 
+import {
+  ExpansionPanel,
+  ExpansionPanelList,
+  ExpansionPanelListItem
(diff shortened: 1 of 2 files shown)


zionts:
Description: An update to the operation registry docs to (hopefully) point out that nothing changes in setting up the operation registry safelisting plugin & options and also a note on guidance that the best practice when running with federation is to always have the operation plugin at the gateway level and to not publicly expose GraphQL traffic from the services.
Assignee: Evans
Reviewer: Trevor

docs/source/platform/operation-registry.mdx
```diff
@@ -169,6 +173,46 @@ const server = new ApolloServer({
...
+
+  server.listen().then(({ url }) => {
+    console.log(`🚀 Server ready at ${url}`);
+  });
+})();
+```
+
+</ExpansionPanel>
+
 ####5.1 Optionally, set the schema tag
```
trevorblades: There should be a space between the last `#` and the number `5` for the heading to render correctly

PR #508, opened 2019-07-05
op-reg: document lifecycle hooks [AP-562]

evans:
This add documentation for the new lifecycle hooks

* willUpdateManifest(newManifest?: OperationManifest, oldManifest?:
OperationManifest) => OperationManifest
  * Run with newManfifest=resultFromGCS when new manifest returned
  from GCS(polled every minute)
  * Run with newManifest=undefined if the manifest is not
  retrieved from gcs on startup
  * oldManifest set to the existing manifest if one is present
* onUnregisteredOperation(requestContext)
  * run when an operation is not found in the safelist
* onForbiddenOperation(requestContext)
  * run when an operation is not in the safelist and
  user-defined override
  (forbidUnregisteredOperations(requestContext) =>
  boolean) doesn't accept the operation

### DIFF
--- a/graph-manager-docs/source/operation-registry.mdx
+++ b/graph-manager-docs/source/operation-registry.mdx
@@ -466,7 +466,7 @@ If a problem occurs during the `apollo service:push` command, make sure that the
 
 Additionally, make sure that introspection is enabled on the server since introspection details are used to obtain and publish the schema.
 
-## Migrating from `0.1-alpha.4` to `0.2.0-alpha.1`
+## Migrating from `0.1-alpha.4` to `0.2.x-alpha.x`
 
 ### Summary of changes
 
--- a/graph-manager-docs/source/operation-registry.mdx
+++ b/graph-manager-docs/source/operation-registry.mdx
@@ -326,7 +326,7 @@ The operation registry plugin supports lifecycle methods to improve observabilit
 > Note: this function is **experimental**, since the manifest shape may change
 
 This **experimental** function is run with `newManifest` set to the result returned by polling the source.
-If the manifest is not retrieved on startup, `willUpdateManfiest` is run with `newManifest` set to `undefined`.
+If the manifest is not available on startup due to an outage of Google Cloud Storage, `newManifest` will be `undefined`.
 After the first call, `oldManifest` is set to the existing manifest if one is present.
 
 This **experimental** function can be used to monitor the number of operations in the manifest.
@@ -344,7 +344,6 @@ const server = new ApolloServer({
           newManifest.operations && 
           newManifest.operations.length
         );
-        return newManifest;
       }
     })
   ]
--- a/graph-manager-docs/source/operation-registry.mdx
+++ b/graph-manager-docs/source/operation-registry.mdx
@@ -323,13 +323,13 @@ The operation registry plugin supports lifecycle methods to improve observabilit
 
 #### willUpdateManifest(newManifest?: OperationManifest, oldManifest?: OperationManifest) => void
 
-**Experimental (manifest shape may change)**
+> Note: this function is **experimental**, since the manifest shape may change
 
-This function is run with `newManifest` set to the result returned by polling the source.
+This **experimental** function is run with `newManifest` set to the result returned by polling the source.
 If the manifest is not retrieved on startup, `willUpdateManfiest` is run with `newManifest` set to `undefined`.
 After the first call, `oldManifest` is set to the existing manifest if one is present.
 
-This function can be used to monitor the number of operations in the manifest.
+This **experimental** function can be used to monitor the number of operations in the manifest.
 
 <ExpansionPanel title="Code sample">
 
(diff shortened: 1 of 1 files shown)


docs/source/platform/operation-registry.mdx
```diff
@@ -273,6 +275,81 @@ const server = new ApolloServer({
 
 > _Note:_ The `forbidUnregisteredOperations` callback must be synchronous. If it is necessary to make an `async` request (e.g. a database inquiry) to make a determination about access, such a lookup should occur within the [`context` function](https://www.apollographql.com/docs/apollo-server/api/apollo-server/#constructor-options-lt-ApolloServer-gt) on the `ApolloServer` constructor (or any life-cycle event which has access to `context`) and the result will be available on the `context` of `forbidUnregisteredOperations`.
 
+### Lifecycle methods
+
+The operation registry plugin supports three lifecycle methods to improve observability and modify the behavior of safelist: `willUpdateManifest`, `onUnregisteredOperation`, and `onForbiddenOperation`.
```
zionts: nit: maybe take this number out in case it becomes out of date in the future?

docs/source/platform/operation-registry.mdx
```diff
@@ -273,6 +275,81 @@ const server = new ApolloServer({
...
+
+#### willUpdateManifest(newManifest?: OperationManifest, oldManifest?: OperationManifest) => OperationManifest
+
+This function is run with `newManfifest` set to the result from gcs, when new manifest is returned by the poll.
+If the manifest is not retrieved from gcs on startup, `willUpdateManfiest` is run with `newManifest` set to `undefined`.
```
zionts: nit: is it important that this is from GCS or can we say "is not retrieved on startup" instead?

docs/source/platform/operation-registry.mdx
```diff
@@ -273,6 +275,81 @@ const server = new ApolloServer({
...
+Every time the function is called `oldManifest` set to the existing manifest if one is present.
```
zionts: This sentence could do with some punctuation, I think?

zionts: Also seems like it may be missing a new line.

docs/source/platform/operation-registry.mdx
```diff
@@ -273,6 +275,81 @@ const server = new ApolloServer({
...
+    })
+  ]
+});
+```
+
+</ExpansionPanel>
+
+#### onUnregisteredOperation(requestContext)
+
+This function is run when an operation is not found in the safelist.
```
zionts: weird that we're calling it the safelist instead of the manifest or registry. I see it as a fact of present implementation that those two are the same.

docs/source/platform/operation-registry.mdx
```diff
@@ -273,6 +275,81 @@ const server = new ApolloServer({
...
+
+<ExpansionPanel title="Code sample">
+
+```js{4-6}
+const server = new ApolloServer({
+  plugins: [
+    require("apollo-server-plugin-operation-registry")({
+      onUnregisteredOperation(requestContext) {
+        metrics.report("safelist.numberUnregisteredOperations", 1);
```
zionts: nit: metrics.increment instead?

docs/source/platform/operation-registry.mdx
```diff
@@ -273,6 +275,81 @@ const server = new ApolloServer({
...
+});
+```
+
+</ExpansionPanel>
+
+#### onForbiddenOperation(requestContext)
+
+This function is run when an operation is not in the safelist and
+`forbidUnregisteredOperations: boolean | (requestContext) => boolean` is
+`true`, ie doesn't accept the operation.
```
zionts: this sentence (and the copies above) are a little inaccurate. It's not when `forbidUnregisteredOperation` is `true`, but when it's either `true` or the predicate returns true for the current request context. Not sure if it's better to be precise here, since wording is probably difficult to avoid clunkiness, but I wanted to point out the inaccuracy.

docs/source/platform/operation-registry.mdx
```diff
@@ -273,6 +275,81 @@ const server = new ApolloServer({
...
+
+<ExpansionPanel title="Code sample">
+
+```js{5-7}
+const server = new ApolloServer({
+  plugins: [
+    require("apollo-server-plugin-operation-registry")({
+      forbidUnregisterdOperations: true,
+      onForbiddenOperation(requestContext) {
+        metrics.report("safelist.numberForbiddenOperations", 1);
```
zionts: nit: metrics.increment instead

docs/source/platform/operation-registry.mdx
```diff
@@ -428,7 +505,7 @@ The new model removes this coupling by using the Apollo api key as the shared
 secret. In order to enable independent secret rotation, the api key is used
 to reference a storage secret that then references the manifests
 
-`/(service id)/storage-secret/(hash of api key).json` <br>
+`/(service id)/storage-secret/(hash of api key).json` <br/>
```
zionts: is this a bug fix?

PR #517, opened 2019-07-13
Fix: Update production document to match the text of Apollo engine website

Lawton:
Going through the tutorial I noticed that the steps to create a new service in Apollo engine didn't reflect the text on the actual page. Perhaps this changed recently on the webpage. Here is a small update for the text to reflect what is actually on Apollo Engine webpage.

Also clarified that the button can show on the sidebar as well. It did for me prior to making a new service/graph.

# TOP
Screen Shot 2019-07-13 at 5 07 56 PM

# CREATE SCREEN
Screen Shot 2019-07-13 at 5 23 12 PM


Lawton:
I also noticed that #444 removes the deploy section, but it is still referenced at the top of the tutorial. Perhaps this should be removed as well?

PR #535, opened 2019-07-29
Glossary: add entry for operation signature

evans:
This PR adds operation signature, explaining that it can be normalized or executable. Additionally we explain what normalization is


docs/source/resources/graphql-glossary.md
```diff
@@ -256,6 +256,32 @@ query getHuman {
...
+```
+query getHuman {
+  human(id: 200) {
+    pounds: weight(unit: "pounds")
+    height
+  }
+}
+```
+
+The normalized operation signature:
```
zionts: this should probably remove whitespace

PR #541, opened 2019-08-05
Remove all references to Zeit Now

Lawton:
In #444 the section on deploying the app was removed, but the section on signing up for Zeit was missed. This means people following the tutorial will create an account for no reason. This PR removes that reference.


zionts:
Thanks for this fix!

PR #559, opened 2019-08-21
Fix named import in code snippet

TylerJBrown192:


StephenBarlow:
Thanks for the catch @TylerJBrown192!

PR #580, opened 2019-09-10
[tutorial 3] resolvers.md: add actual field instead of comment

ryota-murakami:
# Overview

I changed comment to actual type definition in the [3. Write your graph's resolvers](https://www.apollographql.com/docs/tutorial/resolvers/#write-resolvers-on-types) `src/resolvers.js` code highlight.

Full definition appears at [1. Build a schema: Object & scalar types section](https://www.apollographql.com/docs/tutorial/schema/#object--scalar-types), and `type Mission` has 2 field.

- Full definition
actual

Other place, `type Mission` appears second time in [3. Write your graph's resolvers](https://www.apollographql.com/docs/tutorial/resolvers/#write-resolvers-on-types) `src/resolvers.js` code highlight.

- My changed point
mychanged

I copied second one and overwrite full definition, and then I've just noticed lacking `name: String` field by error message when I proceeding [Chapter 5](https://www.apollographql.com/docs/tutorial/client/#make-your-first-query)😅

I thought abbreviation comment may not need for only 1 field.

Thank you for making so practical tutorials! 😄

### DIFF
--- a/docs/source/tutorial/resolvers.md
+++ b/docs/source/tutorial/resolvers.md
@@ -256,7 +256,7 @@ Mission: {
 _src/schema.js_
 ```js
   type Mission {
-    # ... with rest of schema
+    name: String
     missionPatch(mission: String, size: PatchSize): String
   }
 ```


stephtr:
In that case I would highlight the `missionPatch` line. By the way, (why) is the `mission: String` argument needed? The resolver only uses the `size` argument.

ryota-murakami:
@stephtr 
> In that case I would highlight the missionPatch line.

I got what is your intension, looks good!

> By the way, (why) is the mission: String argument needed? The resolver only uses the size argument.

It's my fault, that argument is added more after chapter.

Thank you for make us brilliant material! 🤗

PR #602, opened 2019-09-23
Modify Evans, Danielle, and Adam to code-owners of docs

zionts:
Only own actual docs in `graph-manager-docs` rather than every file in the repo


zionts:
Similarly, happy to change this to be `*.mdx` and `*.md`, though maybe changing images should also have a code owner!
Also happy to delete the whole codeowners thing, but I think it's been helpful.

trevorblades:
@zionts LGTM!

PR #609, opened 2019-09-25
Update breaking changes: union & interface changes

zionts:
This commit fixes up a misrepresentation in our documentation. The `service:check` algorithm counts types added to interfaces and unions as breaking changes. We may revisit this decision in the future, but this commit simply fixes the currently incorrect docs.


zionts:
These rules are actually now valid with changes to the BE.

PR #618, opened 2019-10-01
More explicit instructions

skube:
While I understand this was already in the copy, it still might make things go more smoothly when starting out. Just a suggestion 😄


StephenBarlow:
Thanks for the PR @skube !

PR #619, opened 2019-10-01
Should be clearer where to add other resolvers

skube:


StephenBarlow:
Thanks for the PR! I've clarified this as part of #879.

Issue #239, opened 2018-11-12
Error publishing schema when following tutorial


adokce:
When running `npx apollo schema:check && npx apollo schema:publish` as specified in https://www.apollographql.com/docs/tutorial/production.html#publish I get the following error:
![image](https://user-images.githubusercontent.com/20118440/48354951-35ae5a80-e693-11e8-9f9d-e382616573b1.png)

I followed the tutorial, my server was running in another terminal. I solved the problem by installing apollo and executing a command suggested by Apollo Engine website after creating a service

`apollo schema:publish --endpoint=http://localhost:4000 --key="service:<my-key>"`

At this point I compared my repo and finished repo. There seems to be a difference.
In the final server's index.js I found 

```
const internalEngineDemo = require('./engine-demo');
...
// and inside server setup
engine: {
    apiKey: process.env.ENGINE_API_KEY,
    ...internalEngineDemo,
  },
```

It may be related. I am not 100% sure why I didn't see these lines of code added or explained in tutorial.

Otherwise, thank you for this overall nice tutorial

osteel:
Same here, had to follow Apollo Engine's suggestion:
```
$ npm i -g apollo
$ apollo schema:publish --endpoint=http://localhost:4000 --key="service:<my-key>"
```
`npx apollo schema:check && npx apollo schema:publish` produced the same result as in the comment above.

yuwenhuang:
have the same problem, but I still get `No service found to link to Engine` after I add the following part in `index.js` and manually add `engine-demos.js`

```
const internalEngineDemo = require('./engine-demo');
...
// and inside server setup
engine: {
    apiKey: process.env.ENGINE_API_KEY,
    ...internalEngineDemo,
},
```
hopefully you can help to add more details to this deploy part of tutorial for newbies. I also get error 
`The content of "~/.now/auth.json" is invalid. No `token` property found inside`
after I follow the deploy instruction `npx now`

maxgherman:
It's the **apollo shema:check** that gives this problem. The tutorial command has both, schema:check and schema:publish in it. This disguise a couple of problems:
First, local set up has to be explained. To publish a schema apollo needs 3 things: key, working endpoint and a service name. To set up the key, add it to the index.js

```js
engine: {
    apiKey: process.env.ENGINE_API_KEY
},
```  

Set up env variable  ENGINE_API_KEY for both consoles, the server and the publishing command.
Endpoint is given as part of the command line argument.
[...]

PR #621, opened 2019-10-02
lots of people get problem at this stage see below

aripovula:
pls see 
https://github.com/apollographql/apollo/issues/239


StephenBarlow:
Thanks for the PR! The tutorial now uses a different mechanism for schema registration that has removed this content entirely.

PR #628, opened 2019-10-03
Add reference to Apollo Server proxy configuration documentation.

abernix:
In some environments where outbound traffic is restricted, it may be necessary to configure the Apollo Gateway and Apollo Server to route their outgoing traffic through a proxy (corporate, or otherwise).  These instructions now exist within the Apollo Server documentation (https://www.apollographql.com/docs/apollo-server/proxy-configuration/) an this adds a reference accordingly.

Follows-up: https://github.com/apollographql/apollo-server/pull/3332 (on the Apollo Server repository)

### DIFF
--- a/graph-manager-docs/source/federation.mdx
+++ b/graph-manager-docs/source/federation.mdx
@@ -117,7 +117,7 @@ server.listen().then(({ url }) => {
 });
 ```
 
-> The managed configuration will default to using the `'current'` variant.
+> The managed configuration will default to using the `'current'` variant.  When running Apollo Gateway in an environment where outbound traffic to the internet is restricted, consult the [directions for configuring a proxy](https://www.apollographql.com/docs/apollo-server/proxy-configuration/) within Apollo Server.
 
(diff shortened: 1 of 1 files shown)


abernix:
@StephenBarlow:

@jhampton had a good suggestion in https://github.com/apollographql/apollo-server/pull/3332#issuecomment-534834033, though this doesn't add either of the references he suggested, so there's some opportunity to improve further.

I chose this particular location to provide the reference because it was the first place that the Apollo Gateway was configured in a way which enabled it to talk to our infrastructure (through the setting of `ENGINE_API_KEY`).

It's worth highlighting one potential point of confusion though, since this (existing) section (which I amended) is _above_ the section about _Managed federation_, even though it suggests a mode of enabling managed federation itself.  Perhaps this needs some adjusting/shuffling? 🤹‍♀

StephenBarlow:
Thanks @abernix! I'll go ahead and merge this for now and review more thoroughly as part of a future server content audit.

PR #629, opened 2019-10-03
Clarify options for setting of the Apollo Gateway API key

abernix:
While it is possible for the `ENGINE_API_KEY` to be used, it's also possible to leverage the existing pattern supported by Apollo Server 2.x where the `engine` property can be set to the key, or the more customizable [`EngineReportingOptions`](https://www.apollographql.com/docs/apollo-server/api/apollo-server/#enginereportingoptions).  

Put another way, Apollo Gateway is automatically passed the options from Apollo Server's `engine` options, when they are set.

### DIFF
--- a/graph-manager-docs/source/federation.mdx
+++ b/graph-manager-docs/source/federation.mdx
@@ -96,7 +96,11 @@ Federation is made up of two parts: federated services and a **gateway** to comp
 
 While running in development, it's sufficient to run the gateway with a static service list and use introspection to support the gateway's configuration. When running a gateway in production, however, it's important that the uptime of your gateway is not impacted by the uptime of the federated services beneath it. If a gateway fails to introspect a service or if a service fails to compose with the graph, it's paramount that your gateway be resilient to that change and continue to serve traffic with its previous configuration.
 
-For this use case, an Apollo gateway can be created to pick up its configuration from a set of managed files, securely stored during [service registration](#registering-federated-services). To use the managed configuration provisioned by the Graph Manager, define `ENGINE_API_KEY` in the environment, and construct the gateway and ApolloServer like so:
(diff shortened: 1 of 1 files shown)


graph-manager-docs/source/federation.mdx
```diff
@@ -96,7 +96,11 @@ Federation is made up of two parts: federated services and a **gateway** to comp
...
 
-For this use case, an Apollo gateway can be created to pick up its configuration from a set of managed files, securely stored during [service registration](#registering-federated-services). To use the managed configuration provisioned by the Graph Manager, define `ENGINE_API_KEY` in the environment, and construct the gateway and ApolloServer like so:
+For this use case, an Apollo gateway can be created to pick up its configuration from a set of managed files, securely stored during [service registration](#registering-federated-services).  To use the managed configuration provisioned by the Graph Manager, obtain an API key from the _Settings_ page of the [Graph Manager](https://engine.apollographql.com) and set the `ENGINE_API_KEY` environment variable to that key.
```
abernix: This existing sentence could maybe use some tweaking as well.  Perhaps, this?:

```suggestion
This resilience can be enabled by configuring the gateway in a managed mode where its configuration is securely stored during [service registration](#registering-federated-services), fetched during gateway initialization and then monitored for changes.  To use the managed configuration provisioned by the Graph Manager, obtain an API key from the _Settings_ page of the [Graph Manager](https://engine.apollographql.com) and set the `ENGINE_API_KEY` environment variable to that key.
```

StephenBarlow: I'll give this article a more comprehensive review shortly. In the meantime, I think we can leave this sentence as-is (minus the changes already in this PR of course!)

PR #635, opened 2019-10-05
Inconsistent use of semi-colons

JEStaubach:
Sample code does not have trailing semi-colons which is inconsistent with other examples.


StephenBarlow:
Thanks for the PR @JEStaubach !

PR #639, opened 2019-10-09
Update localSchemaFile docs in "configuring apollo"

JakeDawkins:
The apollo config now supports using multiple `localSchemaFile`s in the config file! 

This PR updates the docs for configuring apollo projects to reflect that. 

Note: the name is purposefully `localSchemaFile` and not `localSchemaFiles` to support backwards-compatibility and to avoid adding yet another option and CLI flag.


JakeDawkins:
@StephenBarlow for things like this, how should we go about documenting or noting that this won't work with older versions of the config used in old CLI releases or old VS code releases?

We added this support in a minor version of both the extension and CLI (`apollo@2.18.2`, `vscode-apollo@1.10.2`)

docs/source/references/apollo-config.md
```diff
@@ -82,19 +82,21 @@ module.exports = {
...
     service: {
       name: 'my-service-name',
-      localSchemaFile: './path/to/schema.graphql'
+      localSchemaFile: ['./path/to/schema.graphql', './another/schema.graphql']
     }
   }
 };
 ```
 
+If you're loading a single schema file, you can also pass the path as a string rather than a list of strings.
```
StephenBarlow: Cut "also"

"rather than" -> "instead of as"

docs/source/references/apollo-config.md
```diff
@@ -82,19 +82,21 @@ module.exports = {
 
 #### _Option 3_: Link a schema from a local file
 
-In some cases you may have a locally generated file with your schema that you want to link. This can be either a `.graphql` file with the schema in SDL form or a saved introspection result in `.json`. To link your client project to a local schema file, configure it like so:
+In some cases you may have one or more locally generated files with your schema that you want to link. These can be either a `.graphql` file with the schema in SDL form or a saved introspection result in `.json`. To link your client project to one or more local schema files, configure it like so:
```
StephenBarlow: "may" -> "might"

"a .graphql file" -> ".graphql files"

"a saved introspection result" -> "saved introspection results"

docs/source/references/apollo-config.md
```diff
@@ -218,12 +220,14 @@ module.exports = {
...
 ```js
 module.exports = {
   service: {
-    localSchemaFile: './path/to/schema.graphql'
+    localSchemaFile: ['./path/to/schema.graphql', './another/schema.graphql']
   }
 };
 ```
+
+If you're loading a single schema file, you can also pass the path as a string rather than a list of strings. If you have many schema files across your project, you can dynamically populate the `localSchemaFile` option using utilities like `glob` to load a list of filepaths.
```
StephenBarlow: See corresponding comment above

docs/source/references/apollo-config.md
```diff
@@ -218,12 +220,14 @@ module.exports = {
 
 #### Option 2: Local schema
 
-In some cases you may have a locally generated file with your schema that you want to link. This can be either a `.graphql` file with the schema in SDL form or a saved introspection result in `.json`. To link your client project to a local schema file, configure it like so:
+In some cases you may have one or more locally generated files with your schema that you want to link. These can be either a `.graphql` file with the schema in SDL form or a saved introspection result in `.json`. To link your server project to one or more local schema files, configure it like so:
```
StephenBarlow: See corresponding comment above

StephenBarlow:
@JakeDawkins Regarding version availability, I recommend a little blockquote notice immediately after the feature in question is mentioned, such as:

```
> Support for linking _multiple_ local schema files is available in the Apollo CLI version 2.18.2 and later, and in the Apollo VS Code extension version 1.10.2 and later. 
```

After enough time has passed that most folks are probably on these versions or later, we can remove the notice.

GuillaumeDesforges:
Had to dig a lot to find this. Hope it will be merged soon.

PR #673, opened 2019-11-08
Typo?

verheyenkoen:
Not sure, but this looks like that should be "Mutation" instead of "Mission".


StephenBarlow:
@verheyenkoen Thanks for the PR! I believe this language is correct as-is, but I agree that it is confusing. This article is overdue for a more significant edit, and I'll add that to my backlog!

PR #677, opened 2019-11-09
Minor fix in the tutorial.

bring-shrubbery:
The file that is asked to be created already exists in the template project.


StephenBarlow:
Thanks for the PR! This change is included in #962.

PR #678, opened 2019-11-09
Removed duplicate query from setup-analytics.md

lorensr:


StephenBarlow:
At some point in the intervening year and a half, this was resolved 😄

PR #684, opened 2019-11-13
Change "cancelTrips" to "cancelTrip" on line 391

csoto09:
The resolver referenced in this line comes from the schema.js file where it is written as the singular _cancelTrip_, rather than the plural _cancelTrips_


StephenBarlow:
Thanks @csoto09 !

PR #685, opened 2019-11-14
Edit to client awareness

StephenBarlow:


graph-manager-docs/source/client-awareness.md
```diff
@@ -1,23 +1,26 @@
...
+
+* `apollographql-client-name`
+* `apollographql-client-version`
 
-Apollo Server 2.2.3+ will look for specific request headers, `apollographql-client-name` and `apollographql-client-version`, by default. If present, Apollo Server will extract them and make sure the data for that request is reported to Apollo's systems with the correct client and version tag.
+If one or both of these headers are present, Apollo Server automatically extracts their values and includes them in the trace report that it sends to Graph Manager. You can [override this default behavior](#advanced-server-configuration).
 
-With Apollo Client 2.4.6+, simply passing the `name` and `version` options in your `ApolloClient` constructor will automatically add these headers to every request. Setting up client identity reporting is as simple as adding configuration to Apollo Client:
+If you're using Apollo Client, you can provide the `name` and `version` options to the `ApolloClient` constructor. If you do, Apollo Client automatically assigns the provided values to the corresponding headers in every operation request.
```
zionts: It would be nice if we could have the information about ApolloClient usage earlier so that users don't need to read all of the implementation details. It would also be nice if there was built-in support for Apollo iOS and others (and there might be!)

Maybe in lieu of that, we could **bold** Apollo Client to call attention to it for AC users?

graph-manager-docs/source/client-awareness.md
```diff
@@ -112,3 +63,28 @@ server.listen().then(({ url }) => {
   console.log(`🚀  Server ready at ${url}`);
 });
 ```
+
+### Using another GraphQL server
```
zionts: Should we have a section on clients with built-in support?

PR #692, opened 2019-11-19
Fix typos in Graph Manager schema validation docs

solon:


StephenBarlow:
Thanks for the PR @solon! These typos have either since been fixed, or the related text has been removed entirely.

PR #694, opened 2019-11-20
Please document local resolvers using the { client } method

breathe:
rather than the { cache } method

To my understanding - method calls like cache.writeData don't broadcast changes to any subscribed components ... why would this be the default used in documentation??  It's ridiculously confusing and very difficult to figure out that what you _actually want_ in 99% of cases is to use client.writeData(...)


StephenBarlow:
Thanks for the PR @breathe! This chapter has since been updated to use new local state APIs in Apollo Client 3.

PR #709, opened 2019-11-27
Operation name `login` should be `Login` in PascalCase

BerlinChan:
Fix graphQL operation name `login`, which should be `Login` with PascalCase


StephenBarlow:
Thanks for the PR! This issue was resolved by other content edits to the article.

PR #712, opened 2019-12-02
Update local-state.md

jonrandahl:
When trying to build I was alerted with `  Line 28:  'GET_CART_ITEMS' is not defined  no-undef` by the compiler. When I checked the final version, the `ggl` import that was specified was already in the code but the `GET_CART_ITEMS` import was missing so I've swapped the imports on line 175 in the snippet above.


StephenBarlow:
Thanks for the PR @jonrandahl! This issue should be resolved in the latest version of the chapter.

PR #715, opened 2019-12-05
Update queries.md

AvneeshD:
Expanding the acronym "AST" to "Abstract Syntax Tree" because it isn't obvious to self taught developers. I am one.


StephenBarlow:
Thanks @AvneeshD! I've sidestepped this issue by removing the terminology entirely.

PR #716, opened 2019-12-05
Fix missing import statement

mjstromberg:
The Launch resolver references a variable that has yet to be imported.


StephenBarlow:
Thanks for the PR @mjstromberg! This should be resolved in the latest version of the chapter.

PR #724, opened 2019-12-11
Convert Full-stack tutorial code samples to TypeScript 

stemmlerjs:
I've gone ahead and rewritten the client-side code blocks in the full-stack tutorial to use TypeScript. 

I'm also making use of the `MultiCodeBlock` component so that code samples can be viewed in both TypeScript and JavaScript.

## Related Open Issue Checklist
Before we merge this, there are a few issues that we would want to resolve first.

- [x] Enable code blocks to use titles.
- [ ] (optional) Fix syntax highlighting on TypeScript blocks with Generics https://github.com/apollographql/gatsby-theme-apollo/issues/77
- [x] Fix TypeScript blocks with unused imports that get removed when switching to JS from TS https://github.com/apollographql/gatsby-theme-apollo/issues/79

## QA Checklist

I'm going to walk through the client-side part of the tutorial again with the new code to make sure that things work and are consistent with the instructions in the docs. 

- [ ] @stemmlerjs did manual QA on the tutorial.


docs/source/tutorial/client.mdx
```diff
@@ -56,49 +58,59 @@ Great, we're all set up! Let's dive into building our first client.
 
 ## Create an Apollo Client
 
+> **The code blocks below use TypeScript by default.** You can use the dropdown menu above a code block to switch to JavaScript.
+>
+>If you're using JavaScript, use `.js` and `.jsx` file extensions wherever `.ts` and `.tsx` are mentioned.
```
trevorblades: 🎉

docs/source/tutorial/schema.mdx
```diff
@@ -168,12 +168,12 @@ Now that we have scoped out our app's schema, let's run the server by calling `s
 _src/index.js_
```
trevorblades: The following code block was converted to TypeScript, but this filename and the codeblock language tag stayed the same. These should probably be changed, and maybe we should wrap the codeblock in a `MultiCodeBlock`.

trevorblades: Alternatively, we can just return this codeblock to its original JS format, since we're not converting the server part of the tutorial to TS yet!

docs/source/tutorial/queries.mdx
```diff
@@ -15,18 +17,25 @@ The `useQuery` hook leverages React's [Hooks API](https://reactjs.org/docs/hooks
 
 ## Fetching a list
 
+> **The code blocks below use TypeScript by default.** You can use the dropdown menu above a code block to switch to JavaScript.
```
trevorblades: Since we're using this same bit of content at the top of each of the client tutorial pages, we could modularize it using MDX and import it in each so that we don't write the same thing 4 times.

The pattern we've been using for this is to create an MDX file in the **docs/shared** directory (or any sibling of **source**), paste the shared content in there, and then render it as if it were a JSX component.

_docs/shared/disclaimer.mdx_

```markdown
**The code blocks below use TypeScript by default.** You can use the dropdown menu above a code block to switch to JavaScript.

If you're using JavaScript, use `.js` and `.jsx` file extensions wherever `.ts` and `.tsx` are mentioned.
```

_docs/source/tutorial/queries.mdx_

```jsx
import Disclaimer from '../../shared/disclaimer.mdx';

<Disclaimer />
```

docs/source/tutorial/schema.mdx
```diff
@@ -187,13 +187,13 @@ By default, Apollo Server supports [GraphQL Playground](https://www.apollographq
 The GraphQL Playground provides the ability to introspect your schema. **Introspection** is a technique used to provide detailed information about a graph's schema. To see this in action, check out the right hand side of GraphQL Playground and click on the `schema` button.
 
 <div style="text-align:center">
-  <img src="../images/schematab.png" alt="Schema button">
```
trevorblades: We could probably undo all of the changes to this file and turn it back into a regular old MD file to reduce the footprint of this PR. This should be straightforward to do using git, and I'd be happy to help do that if you agree with the change!

stemmlerjs: I agree, yeah @trevorblades would you mind reverting it for me? Thanks!

trevorblades: Done! 🎉

PR #734, opened 2019-12-18
Change LaunchConnection to LaunchCollection spelling fix in documentation

raheelkhan:
the paginated list of launches should be LaunchCollection instead of LaunchConnection


StephenBarlow:
Thanks for the PR @raheelkhan! `LaunchConnection` is correct here, because the return type represents a "connection" to a paginated field that is queried multiple times.

PR #736, opened 2019-12-24
Update resolvers tutorial highlighting

stephtr:
"Authenticate users" is the first time the `context` is added.


StephenBarlow:
Thanks for the PR! I've incorporated this change as part of #879

PR #737, opened 2019-12-24
Client Tutorial: consistently use single quotes

stephtr:
Change double to single quotes, as being used for all the other imports in the sample code.


StephenBarlow:
Thanks @stephtr! I've merged https://github.com/apollographql/apollo/pull/850, which also resolves this.

PR #743, opened 2020-01-01
suggested changes

sbnsal:
We updated the dataSource property in ApolloServer, right ?


StephenBarlow:
Thanks for the PR! I've updated the language here with #879

PR #744, opened 2020-01-05
Update resolvers.md

rafmjr:
I propose removing unnecessary async keyword for regular function.

I also found confusing the code block for src/schema.js following the explanation of the resolver for Mission type. The schema shown in the code block differs from what was shown to the user in previous sections and the changes introduced are not relevant as the mission argument that precedes size is never used on this section.

### DIFF
--- a/docs/source/tutorial/resolvers.md
+++ b/docs/source/tutorial/resolvers.md
@@ -253,14 +253,6 @@ Mission: {
 },
 ```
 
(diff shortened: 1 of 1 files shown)


rafmjr:
Done! Thank you very much for your quick feedback.

rafmjr:
Done! Thank you very much for your quick feedback.

PR #747, opened 2020-01-07
Update _redirects to fix apollo-config reference

JakeDawkins:
Fixes https://github.com/apollographql/apollo-tooling/issues/1195


JakeDawkins:
The redirect from `/docs/references/apollo-config.html` works with this link in the preview: https://deploy-preview-747--apollo-docs-index.netlify.com/docs/references/apollo-config.html

PR #780, opened 2020-02-08
Remove vestigial(?) reference to previous section

jaimemarijke:
I'm not sure if this is trying to say that we cloned the entire repo in the previous section, or just that we copied code into the `src/utils.js` file in the previous section, but either way that isn't true. We cloned the repo several sections ago, and we didn't have to explicitly copy the `utils.js` file ever. Seems more concise to just remove it.


StephenBarlow:
@jaimemarijke Thanks for the PR!

PR #788, opened 2020-02-18
A incorrect file type fixed(small issue)

thomasJs8:
## Summary
In a tutorial of `6. Fetch data with queries`, a file type is written `jsx` instead of `tsx`.  

## issue
The file name is `src/pages/launch.tsx`, but that file type is `JavaScript`.
I think it should be fixed.

#### Incorrect
[Screenshot from Gyazo](https://gyazo.com/9edd8937f8d5f00f0edbbc47481bb3eb)

#### Correct
[Screenshot from Gyazo](https://gyazo.com/2ff87d9d1ff8be46993f25e14a60a40d)

Thank you.

### DIFF
--- a/docs/source/tutorial/queries.mdx
+++ b/docs/source/tutorial/queries.mdx
@@ -341,7 +341,7 @@ Let's use our fragment in our launch detail query too. Be sure to import the fra
 
 <MultiCodeBlock>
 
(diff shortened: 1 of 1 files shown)


StephenBarlow:
@thomasJs8 Thanks for the PR!

PR #790, opened 2020-02-22
typing for engine logging not correct

maapteh:
when using corporate proxy it seems the API expects a requestAgent (for example: new HttpsProxyAgent(proxyUrl)) and not just a proxyUrl.


maapteh:
Oeps i now see almost everything is not correct on engine part. Maybe auto-generate this from the TS declarations?

screenshot

abernix:
The page you're referring to is for the deprecated Engine Proxy (a Go binary with a JSON-configuration format), not the Apollo Engine Reporting that is built into Apollo Server.  Perhaps you're looking for [this documentation](https://www.apollographql.com/docs/apollo-server/proxy-configuration/)?

PR #795, opened 2020-02-28
Can this line be removed?

siddtheone:
It's getting added while coping and causing error. May be confusing for readers.


trevorblades:
@siddtheone could you please remove the entire line with the ellipsis, rather than replacing it with an empty line?

siddtheone:
@trevorblades 
Line removed.

StephenBarlow:
@siddtheone Thanks for the PR!

PR #802, opened 2020-02-29
fix: some query exclamation points were missing

EdnilsonCS:
The query exclamations were missing


EdnilsonCS:
I thought there was an error, more researching better I found that the slogan is an optional parameter, to say that the argument is mandatory

PR #806, opened 2020-03-05
fix: keeps code sample consistent

alvincrespo:
## Changes

- Adds `link` const as per https://www.apollographql.com/docs/tutorial/client/


StephenBarlow:
Thanks for the PR @alvincrespo! This chapter now uses the `uri` option for `ApolloClient` in place of `link`

PR #810, opened 2020-03-12
Remove repetition in federation.mdx

lorensr:


StephenBarlow:
Thanks @lorensr ! I've incorporated this change in #977

PR #812, opened 2020-03-16
Update resolvers.md

ionutale:
this part is a bit misleading
for simplicity i added the login resolver  again together

it is not the best solution...


StephenBarlow:
Thanks for the PR! I've clarified this using comments in #879

PR #817, opened 2020-03-21
Update production.md

Yasin-Ali:
Following the documentation led to this error ```Error: No service found to link to Apollo Graph Manager```
The instructions on the Apollo Graph page used `user` instead of `service` in the ENGINE_API_KEY and called for the creation of an Apollo Config file.


zionts:

PR #818, opened 2020-03-23
Changed "Apollo data source" to "Apollo data source API"

shani-asad:
At line 10,I think it should be Apollo data source API, as it is an explanation for the previous sentence. a data source API is different from data source, right?


StephenBarlow:
Thanks for the PR! I've clarified this language in #873.

PR #820, opened 2020-03-25
Getting Started article for Graph Manager

StephenBarlow:


graph-manager-docs/source/getting-started.mdx
```diff
@@ -71,22 +73,24 @@ In Graph Manager, each **graph** corresponds to a different GraphQL schema. Your
     ```js{3}:title=apollo.config.js
     module.exports = {
       service: {
-        name: "YOUR_SERVICE_NAME"
+        name: "YOUR_GRAPH_NAME"
       }
     }
     ```
 
 ## Step 4: Register your schema
```
michael-watson: I think we need to break this out into monolith schema and a federated schema. We have users that register their entire gateway or create a graph in AGM for every service. See https://app.intercom.com/a/apps/lb30fyal/inbox/inbox/unassigned/conversations/26397301185

michael-watson: this comment looks weird, but I mean in the Register your schema section

StephenBarlow: I've resolved this by directing everyone getting started with a federated schema to view the managed federation setup instead of this article (at the very top)

PR #846, opened 2020-04-11
Remove extraneous code snippet

danielmichaelni:
1. The `schema.js` code snippet is irrelevant to this section
2. The `missionPatch` field is incorrect. It should be `missionPatch(size: PatchSize): String` rather than `missionPatch(mission: String, size: PatchSize): String`

### DIFF
--- a/docs/source/tutorial/resolvers.md
+++ b/docs/source/tutorial/resolvers.md
@@ -252,14 +252,6 @@ Mission: {
 },
 ```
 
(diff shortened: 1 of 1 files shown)


StephenBarlow:
Thanks for the PR! With #879, I've clarified that you don't need to copy and paste this code block and I've removed the extraneous argument. We'll keep it around to remind the reader what the mutation's definition looks like.

PR #859, opened 2020-04-24
make iframes responsive

kkemple:
This PR makes iFrames responsive so that they don't overlap the side menu. I feel like this could probably become a component that ships with the docs theme, but I chose not to do that extra work without having more iframes that needed to be fixed.

Before:
<details>
Screen Shot 2020-04-14 at 12 16 58 PM
</details>

After:
<details>
Screen Shot 2020-04-14 at 12 16 58 PM
</details>


trevorblades:
I like this idea! Since we're using this pattern for multiple iframes in the site, what do you think about turning this pattern into a reusable component?

```js
// src/graph-manager-docs/components/video-frame.js
import React from 'react';

export default function VideoFrame(props) {
  return (
    <div
      style={{
        position: 'relative',
        overflow: 'hidden',
        paddingTop: 'calc(100% / (16 / 9))'
      }}
    >
      <iframe
        style={{
          position: 'absolute',
          top: 0,
          left: 0,
          width: '100%',
          height: '100%',
          border: 0
        }}
        src={props.src}
        frameborder="0"
        allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
        allowfullscreen
      />
    </div>
  )
}
```

And then we can use them in the MDX files like this:

```md
import { VideoFrame } from "../components/video-frame";

<VideoFrame src="https://www.youtube.com/embed/yZ3_Yvlmy78" />
```

kkemple:
@trevorblades I always forget about calc! nice addition!

trevorblades:
Yeah!! `calc` ftw 🎉

kkemple:
This is ready for a new review! Made a component, added titles for A11y.

PR #864, opened 2020-04-29
Update tutorial step 4.

KBora:
I was not able to follow the existing instructions as it seems the UI / setup in Apollo Graph Manager has been updated.


zionts:

PR #868, opened 2020-05-04
fix spelling in the client page of the tutorial

mguida22:
Hello! I've been going through the tutorial and noticed what I think is a small spelling error.


StephenBarlow:
Thanks for the catch @mguida22 !

mguida22:
No problem!

PR #869, opened 2020-05-04
Tutorial docs fix: remove `mission` argument from `missionPatch`

### LINKED ISSUE #532: Type bug in the resolvers tutorial example page
There's a type error on the `missionPatch` resolver. The first argument should be a `Mission` type instead of a `String` type

```
type Mission {
    # ... with rest of schema
    missionPatch(mission: String, size: PatchSize): String
  }
```

pcattori:
The resolver *does* take `mission` as its first argument (since the mission is the parent of the mission patch), but the `missionPatch` field in the type definition for `Mission` should not take a `mission` argument.

Fixes #532

### DIFF
--- a/docs/source/tutorial/resolvers.md
+++ b/docs/source/tutorial/resolvers.md
@@ -256,7 +256,7 @@ _src/schema.js_
 ```js
   type Mission {
     # ... with rest of schema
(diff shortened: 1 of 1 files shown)


StephenBarlow:
Thanks for the PR! I've incorporated this change as part of #879

PR #872, opened 2020-05-07
Updated production.md

shubuu97:
Changes:
1-) "ENGINE_API_KEY" is deprecated and new key name is "APOLLO_KEY".
2-) The place where hash key appears is also changed.
3-) Code for Publishing the schema is also updated.


zionts:

PR #876, opened 2020-05-12
Proposed grammar change

farjadfazli:
"cursor at the end of the..." sounds better to me than "cursor of the end of the..." but not sure if that changes the intended meaning. Is the cursor something that _belongs_ to the end of the paginated results or something that exists _at_ the end of the paginated results?

Please advise!


StephenBarlow:
Thanks for the PR @farjadfazli !

PR #877, opened 2020-05-13
Outdated documentation

Glen-Moonpig:
Documentation around environment variables and property names is out of date.

The documentation in general is a bit of a mess and the same concepts in different Apollo packages have taken on different names making it very difficult to reconcile something in one apollo package with another...


zionts:

PR #880, opened 2020-05-15
docs: update client.mdx

MrHeer:
update `step 4` to `step 5`

because https://github.com/apollographql/apollo/pull/879


chipcullen:
FWIW, I made this change, as well as a lot more, in https://github.com/apollographql/apollo/pull/885

StephenBarlow:
Thanks for the PR! This error is resolved with #962.

PR #885, opened 2020-05-28
Update client.mdx

chipcullen:
I'm proposing the following changes to the client tutorial, based on things that tripped me up:

- Explaining what an AST is, albeit briefly
- An extensive re-write of the section on getting the VSCode extension up and running. There were several issues with that section, such as the extension not being named fully (so as to make it clear when searching for the extension in the marketplace). Also, the `space-explorer` service doesn't work anymore (I don't know, but I wonder if it was a publicly available service once upon a time), so I am suggesting people use their own service name (this is what worked for me in local testing anyway). I also tell people to, you know, _install_ the extension. 
- Lastly I explain more fully that they can use `localhost` in the text about setting up the `uri`.


StephenBarlow:
Thanks for this PR! I've merged a separate set of changes with #962 that I believe address all of the issues addressed here.

PR #886, opened 2020-05-28
Update local-state.mdx

chipcullen:
Clearing up a few more stumbling blocks that I had on this page. 

One overall issue that I've noticed - for the codeblocks, the `title` attribute does not seem to work on these typescript examples (though they seem to work fine with the js examples in page 1). So, the helpful hint of which file we're editing isn't there. I've literally had to come to the github file editing interface to track down which file the tutorial is referring to. I've tried to make inline comments to tell people where to look, as it can be very confusing. 

I've also updated the last `export const resolvers` example to match what I needed to do to make the project compile. Mostly, I had to indicate that `Launch` still needed to be there.

### DIFF
--- a/docs/source/tutorial/local-state.mdx
+++ b/docs/source/tutorial/local-state.mdx
@@ -134,7 +134,7 @@ function IsLoggedIn() {
 injectStyles();
 ReactDOM.render(
   <ApolloProvider client={client}>
-    <IsLoggedIn />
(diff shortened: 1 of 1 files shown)


StephenBarlow:
Thanks for the PR @chipcullen! This chapter has now been updated to use new local state APIs in Apollo Client 3, and hopefully the spirit of these changes are included in that update.

PR #890, opened 2020-05-31
[Docs/Tutorial 6]: Configure Apollo VSCode

inglkruiz:
Aligns the client service `id` / `name` placeholder because it was a bit misleading. Adds a new Note in case you receive an error on Loading Schema. This happened to me, I guess due to I installed the plugin before configuring the service in `apollo.config.js`.


StephenBarlow:
Thanks for the PR! This setup is clarified as of #962.

PR #892, opened 2020-06-01
Add schema reporting documentation

StephenBarlow:


graph-manager-docs/source/schema-registry.mdx
```diff
@@ -5,23 +5,71 @@ description: A central hub for your data graph
...
+    ```none:title=.env
+    APOLLO_KEY=YOUR_KEY_HERE
+    ```
+
+3. Set the `reportSchema` option to `true` in your `ApolloServer` constructor, like so:
+
+    ```js
+    const server = new ApolloServer({
+      // ...other options...
+      reportSchema: true,
```
zionts, line 39: We will need to nest this within the `engine: { ... }` key 🙏

graph-manager-docs/source/schema-registry.mdx
```diff
@@ -5,23 +5,71 @@ description: A central hub for your data graph
...
+    });
+    ```
+
+Now every time your server starts up, it registers its schema with Graph Manager.
```
zionts, line 43: Should we explain the nuance here that we wait for deploys to finish or is that left to the more intrepid reader?

graph-manager-docs/source/getting-started.mdx
```diff
@@ -114,7 +114,7 @@ To begin pushing these traces to Graph Manager, read [Pushing trace data to Grap
 
 Your schema is now registered with Graph Manager, and your server is pushing operation traces!
 
-However, to get the most out of Graph Manager, you need to _re_-register your server's schema every time it changes. The easiest way to do this is to make schema registration part of every deployment in your application's continuous delivery flow. [See an example.](schema-registry/#registering-a-schema-via-continuous-delivery)
+However, to get the most out of Graph Manager, you need to _re_-register your server's schema every time it changes. The easiest way to do this is to make schema registration part of every deployment in your application's continuous delivery flow. [See an example.](schema-registry/#registering-with-continuous-delivery)
```
zionts, line 117: we should probably recommend using schema reporting for this instead of pointing to CD

graph-manager-docs/source/schema-registry.mdx
```diff
@@ -5,23 +5,71 @@ description: A central hub for your data graph
...
 
-## Registering a schema manually
+#### Registering automatically to a variant
 
-Register your schema with Apollo Graph Manager by running the `apollo service:push` command from your application's root directory, like so:
+Your server can register its schema to a particular [variant of your graph](#managing-environments-with-variants). Each of your server's environments (development, staging, production, etc.) should register to a different variant.
+
+To do so, set the `APOLLO_GRAPH_VARIANT` environment variable in each server environment:
```
zionts, line 49: Should we also mention you can set this in the `ApolloServer` constructor with the
```
engine: { graphVariant: "prod" }
```
option as well?

graph-manager-docs/source/schema-registry.mdx
```diff
@@ -33,15 +81,27 @@ id      schema        tag
...
-## Registering a schema via continuous delivery
+If you omit the `--variant` flag, the `apollo service:push` command pushes to the default graph variant, named `current`.
 
-To get the most out of Graph Manager, you should **register each update to your production schema as soon as it occurs**. To achieve this, include a call to `apollo service:push` in your continuous delivery pipeline.
+## Registering with continuous delivery
 
-Here is a sample continuous delivery configuration for pushing a schema to Apollo using CircleCI:
+To get the most out of Graph Manager, you should **register each update to your production schema as soon as it occurs**. Consequently, schema registration should be part of your continuous delivery pipeline.
+
+If you're using [automatic schema reporting](#automatic-registration), registration happens automatically each time you deploy your production server. No additional continuous delivery configuration is required.
```
zionts, line 100: should we link to the "tie-breaking" rules here, and explain that the registration will happen at the _end_ of the deploy rather than the beginning?

graph-manager-docs/source/schema-reporting-protocol.mdx
```diff
@@ -0,0 +1,153 @@
...
+https://edge-server-reporting.api.apollographql.com/api/graphql
+```
+
+## Protocol sequence
+
+The following sections illustrate the communication sequence between a GraphQL server (hereafter referred to as the **edge server**) and the schema reporting endpoint (**Apollo**):
+
+### Diagram
+
+```mermaid
```
zionts, line 25: this diagram appears with overlapping text in the preview :( but maybe that shall be fixed soon

graph-manager-docs/source/schema-registry.mdx
zionts: Ahh, I see that section didn't carry over to these docs. I think it's important to document what users should expect to happen when they deploy, since there will be a slight delay in upgrading traffic. It should have had its own section in the preview docs, but it's here: https://github.com/apollographql/apollo-schema-reporting-preview-docs#the-schema-reporting-endpoint

zionts: Do you think it's also worthwhile to note how schema reporting interacts with calls to `service:push` or is it unnecessary to document what we might consider the "expected" behavior?

zionts: One last section that is missing is the additional config option for _which schema_ we send and how to override it: https://github.com/apollographql/apollo-schema-reporting-preview-docs#the-schema-reporting-endpoint

graph-manager-docs/source/getting-started.mdx
StephenBarlow: Agreed. I'd prefer to leave out of this PR, as there is a separate set of changes coming for this article that will bring schema reporting much more front and center.

graph-manager-docs/source/schema-registry.mdx
StephenBarlow: I'd prefer to provide only a single method, and only the environment variable method. The constructor option method feels less appropriate for a value that changes specifically based on environment

graph-manager-docs/source/schema-reporting-protocol.mdx
StephenBarlow: It shall! https://github.com/apollographql/gatsby-theme-apollo/pull/123

graph-manager-docs/source/schema-registry.mdx
StephenBarlow: Whoops! Yep, straight-up left stuff out

graph-manager-docs/source/schema-reporting-protocol.mdx
```diff
@@ -0,0 +1,159 @@
...
+    withSchema = true
+  }
+
+  setTimeout(sendReport, response.inSeconds)
+}
+
+sendReport()
+```
+
+## Type definitions
```
Enrico2, line 87: This section needs an update. See https://github.com/mdg-private/monorepo/blob/master/docs/engine_services/auto_registration/auto_schema_reporting.md

graph-manager-docs/source/schema-reporting-protocol.mdx
```diff
@@ -0,0 +1,159 @@
...
+| Name  | Type | Description  |
+|---|---|---|
+| `bootId` | `String!` | A randomly generated UUID that's unique for each instance of your edge server. Set this value on server startup (a given value should not persist across restarts). |
+| `executableSchemaId` | `String!` | A unique identifier for the edge server's schema. Should be the hex representation of the schema document's SHA-256 hash. |
+
+### Recommended fields
+
+| Name  | Type | Description  |
+|---|---|---|
+| `graphVariant` | `String!` | The name of the graph variant to register the schema to. The default value is `current`. |
```
Enrico2, line 135: Technically this _is_ required in the API, but has a default, so I guess it depends on the question 'when we say required, do we mean "required to provide" or "required to be sent"'.

graph-manager-docs/source/schema-reporting-protocol.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+| Name  | Type | Description  |
+|---|---|---|
+| `runtimeVersion` | `String` | The runtime that your edge server is running, such as `node 12.03`.
+| `libraryVersion` | `String` | The name and version of the server and/or reporting agent your edge server is using, such as `apollo-server-2.8` or `graphql-java-3.1`. | 
+| `platform` | `String` | The infrastructure environment that your edge server is running in (`localhost`, `kubernetes/deployment`, `aws lambda`, `google cloud run`, `google cloud function`, `AWS ECS`, etc.) |
+
+## Schema normalization
+
+Two semantically identical schemas can appear different to Graph Manager, for example if those schemas list the same set of object fields in a different order. To avoid this, every edge server should **normalize** its schema before sending it to Graph Manager.
```
Enrico2, line 151: Are we replacing Graph Manager with Studio already or is that something we're going to do on a defined time?

graph-manager-docs/source/schema-reporting-protocol.mdx
Enrico2: same for the step by step instructions above to match the types.

graph-manager-docs/source/schema-registry.mdx
```diff
@@ -5,23 +5,110 @@ description: A central hub for your data graph
...
+});
+```
+
+### Other GraphQL servers
+
+Any GraphQL server can perform automatic schema registration by implementing the [Schema reporting protocol](./schema-reporting-protocol/).
+
+> Currently, we are not aware of GraphQL server libraries besides Apollo Server that implement the schema reporting protocol. If you add support to an open-source library, please let us know at **support@apollographql.com**!
+
+### Handling conflicting schemas
```
zionts, line 94: I'm worried that users who are trying to answer the question of "what happens when I have multiple servers w/ different schemas?" will find this (but are rare) and those who are wondering "what happens during a deployment of a new schema?" might not. Perhaps we could name this somewhat more generically, like "Breaking ties" or "Handling roll-out of a new schema"?

graph-manager-docs/source/schema-registry.mdx
```diff
@@ -5,23 +5,110 @@ description: A central hub for your data graph
...
+
+Whenever you perform a rolling deployment, different active instances of your server might report different schemas to Graph Manager. When this occurs, Graph Manager uses the following algorithm to choose which schema to register:
 
-## Registering a schema manually
+1. If the specified variant does not yet have an active schema, register the most recently reported schema.
+2. If the specified variant _does_ have an active schema and that exact schema has been reported within the last 120 seconds, keep the current active schema.
```
zionts, line 99: 120 has actually become 90 nao

graph-manager-docs/source/schema-registry.mdx
```diff
@@ -33,15 +120,27 @@ id      schema        tag
...
+apollo service:push --variant=beta
+```
+
+If you omit the `--variant` flag, the `apollo service:push` command pushes to the default graph variant, named `current`.
+
+## Registering with continuous delivery
+
+To get the most out of Graph Manager, you should **register each update to your production schema as soon as it occurs**. Consequently, schema registration should be part of your continuous delivery pipeline.
+
+**If you're using [automatic schema reporting](#automatic-registration)** with Apollo Server, registration happens automatically after your server's deployment completes. No additional continuous delivery configuration is required.
```
zionts, line 139: Can we specifically say this is the preferred method? I always prefer, when specifying multiple options, to point out which is the recommended way

graph-manager-docs/source/schema-reporting-protocol.mdx
StephenBarlow: That will be part of a subsequent PR

graph-manager-docs/source/schema-reporting-protocol.mdx
StephenBarlow: Right, I am erring on the side of organizing by "whether or not your code needs to provide a value" as opposed to "whether or not the field is nullable in the schema"

zionts: There's often a "default" column for use cases like this.

PR #895, opened 2020-06-04
Improved fragment example readability

RickvdP:
Minor change but increased readability for following tutorial - the second changed line was very easy to miss.


StephenBarlow:
Thanks for the PR @RickvdP! This chapter has been updated so that codeblocks are copied and pasted in full.

PR #897, opened 2020-06-05
Initialise cache and link first to maintain consistency from previous steps

misidoro:
In order to maintain consistency from the previous steps, would be good to initialise `cache` and `link` first, before creating a new `ApolloClient`.


StephenBarlow:
Thanks for the PR @misidoro! The cache is now indeed initialized before ApolloClient (in a separate file), and the use of `link` has been removed.

PR #898, opened 2020-06-08
Comma was missing in client object

bansalsohil:


StephenBarlow:
Thanks @bansalsohil !

PR #900, opened 2020-06-15
Update GraphQL Playground Screenshots

caillou:
When going through the tutorial, it took me a moment to figure out that I had to click the docs tab, as the current screenshots suggests to click on `SCHEMA` rather than `DOCS`.


StephenBarlow:
Thanks for the PR! References to Playground have been removed from the tutorial in favor of Apollo Sandbox.

PR #907, opened 2020-06-21
Update production. md

dfrho:
Proposing, and requesting better clarity in these instructions. If I understood them I could provide the proper edit.


dfrho:
Leaving a message on Spectrum channel instead.

PR #910, opened 2020-06-22
Update client.mdx

cseas:
Corrected step reference and add a link to the exact section being referred


StephenBarlow:
Thanks for the PR! The link to the relevant previous step is updated as of #962.

PR #912, opened 2020-06-22
Fix anchor names that error

StephenBarlow:


StephenBarlow:
This error is being caused by an upstream dependency of Gatsby that should be resolved in its next release.

PR #913, opened 2020-06-23
Missing colon in code example

DigitalVilla:


StephenBarlow:
Great catch, thanks @DigitalVilla! I went ahead and incorporated this as #914

PR #920, opened 2020-06-29
Update mutations.mdx

taina0407:
Highlight correct changed code on "Expose Apollo Client with useApolloClient" segment


StephenBarlow:
Thanks @taina0407! I've incorporated this change in #977

PR #922, opened 2020-07-01
docs(schema-reporting): fix missing colon in code configuration examples

Addono:
Without the colons the code examples do not use valid Javascript for declaration of objects.


StephenBarlow:
Thanks for the catch @Addono !

PR #924, opened 2020-07-06
Update client.mdx

Edenghajakk:
I think the correct terms here is `imported` instead of `exported`. Correct me if I'm wrong😃


StephenBarlow:
Thanks for the PR! This language is clarified as of #962.

PR #931, opened 2020-07-12
Don't use 'loading' state in client cache query

srekhi:
Because 'GetCartItems' is a cache read, no need to account for 'loading' state


StephenBarlow:
Thanks for the PR @srekhi! This chapter has been updated to use Apollo Client 3 local state APIs, and no loading states are needlessly used.

PR #939, opened 2020-07-18
docs: mark added lines

andrewvasilchuk:
This change highlights the lines that was added.


StephenBarlow:
Thanks @andrewvasilchuk! I've incorporated this change as part of #977

PR #945, opened 2020-07-23
Fix a broken link

StephenBarlow:


StephenBarlow:
Very confused that this is required, should be handled by changes to `_redirects`...

PR #948, opened 2020-07-26
Add subfields to LoginUser mutation to avoid error

nbhankes:
This mutation query without a subfield results in the following error:     

"errors": [
      {
        "message": "Field \"login\" of type \"User!\" must have a selection of subfields. Did you mean \"login { ... }\"?"

In order to prevent this error, I added two subfields. I maintain use of the original token string and made minor edits to then isolate the token string from the mutation output and continue with the original flow of the tutorial.


StephenBarlow:
Thanks for the catch @nbhankes!

PR #950, opened 2020-07-31
[GM-1392] Recommend schema check configurations over CLI flags 

caydie-tran:
As it is right now, it sounds like we recommend using CLI flags to customize validation. But we should instead make it clear that we recommend controlling this centrally in Apollo Studio with Check Configurations.


StephenBarlow:
Resolved by #954

PR #953, opened 2020-08-02
Update tutorial with latest packages

rooiss:
The current tutorial references old/invalid npm package names.


StephenBarlow:
Thanks for the PR! These changes were incorporated as part of another PR to update the tutorial for AC3.

PR #963, opened 2020-08-14
Fix Apollo Client Getting Started link in FAQ.

xikaos:
The link was: https://www.apollographql.com/docs/react/essentials/get-started.html
Changed to: https://www.apollographql.com/docs/react/v2.5/essentials/get-started/


StephenBarlow:
Thanks for the PR @xikaos ! I've incorporated this change in #977

PR #968, opened 2020-08-18
Add missing param on documentation for successfully connection to Apollo Studio

MussaCharles:
Apollo Studio seems to have some changes when creating new graph,I think there is one extra parameter (graphVariant) which seems to be required to successfully receive reports from Apollo server. However it is not reflected on the current docs.  

This pull request adds the missing parameter so that to avoid others to face the same issue as I had when trying to connect without it.

### DIFF
--- a/docs/source/tutorial/production.md
+++ b/docs/source/tutorial/production.md
@@ -52,7 +52,8 @@ Now, add one extra option to your `ApolloServer` constructor to enable automatic
 const server = new ApolloServer({
   // ...other options...
   engine: {
-    reportSchema: true
+    reportSchema: true,
+    graphVariant: "current"
   }
 });
 ```


StephenBarlow:
Thanks for the PR @MussaCharles! I've double-checked this behavior with the team, and the `graphVariant` field should _not_ be required (defaults to `current`). In our testing, we were able to report a schema without the field.

As to the issue you encountered, we suspect one of the following:

- By default, Apollo Server waits a random amount of time between 0 and 10 seconds after startup before it reports its schema. This is to stagger schema reports if you deploy multiple server instances simultaneously.
- Reported schema changes are queued for processing by Studio before they appear on your end, which can take on the order of a minute.

All of which is to say, you should be able to remove the `graphVariant` field and things will continue to work. If you do remove the field and continue to encounter issues, please at-reply to me here and we will investigate further!

PR #972, opened 2020-08-25
Update data-source.md

vishwaratna:
The dependency has been changed now.


trevor-scheer:
I don't think this is correct. I'm curious to know why you suspect otherwise, though. Is there something confusing in our docs that led you to believe this?

See exports from `apollo-server-caching` here:
https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-caching/src/index.ts

See exports from `apollo-datasource-rest` here:
https://github.com/apollographql/apollo-server/blob/main/packages/apollo-datasource-rest/src/index.ts

PR #978, opened 2020-08-28
Redirect /docs/angular to new home

kamilkisiela:
We moved apollo-angular under my account a month ago


StephenBarlow:
Whoops missed this one when I merged those other link fixes @kamilkisiela

PR #979, opened 2020-08-28
Update resolvers.md

devGuerra:
Fix name LaunchAPI, launchAPI not be resolved


StephenBarlow:
Hi @devGuerra, thanks for the PR! These symbol names are correct as-is, according to the mapping defined in the previous tutorial chapter, here: https://www.apollographql.com/docs/tutorial/data-source/#add-data-sources-to-apollo-server

PR #986, opened 2020-09-02
Fix link to reference implementation for signature

lsanwick:
The current link is a 404, the new link is my best guess at the proper link under the newer utility package.


StephenBarlow:
Thanks for the PR @lsanwick! It looks like this change was handled by another commit that was subsequently merged into this PR.

PR #991, opened 2020-09-09
docs: update queries.mdx, avoid checking data is defined

gaspardip:
We know data is defined here because we have explicitly checked so a few lines above.


StephenBarlow:
Thanks for the PR! You are absolutely correct. For purposes of reducing confusion in case a developer chooses to remove the  `Not found` logic above, we'll leave this as-is for the time being.

PR #996, opened 2020-09-15
Update resolvers.md

idkjs:
need to add `trips: [Launch]!` to `type User`


StephenBarlow:
Thanks for the PR @idkjs! The schema defined in step 1 already defines the `trips` field on the `User` type, so this addition isn't necessary here.

PR #1025, opened 2020-10-19
Mutation chapter updates to follow Queries

StephenBarlow:
Only one chapter to go after this one!

Also, much like the Queries chapter, there are changes to the tutorial codebase that need to accompany these changes.

### DIFF
--- a/docs/source/tutorial/mutations.mdx
+++ b/docs/source/tutorial/mutations.mdx
@@ -11,7 +11,7 @@ Time to accomplish: _10 Minutes_
 
 Now that we've [added multiple queries](./queries/) to our React app, let's add a **mutation**. The process is similar, with a few important differences.
 
-As a reminder, mutations are operations that can modify back-end data (unlike queries). Our server's schema supports the following mutations:
+As a reminder, mutations are GraphQL operations that can modify back-end data (unlike queries). Our server's schema supports the following mutations:
 
 ```graphql
 # YOU DON'T NEED TO COPY THIS.
@@ -109,7 +109,7 @@ const [login, { loading, error }] = useMutation<
 );
 ```
 
-Our `onCompleted` callback stores the user's unique ID and session token in `localStorage`, so we can load these values into the in-memory cache the next time the user visits our application. We'll add that functionality in the next section.
+Our `onCompleted` callback stores the user's unique ID and session token in `localStorage`, so we can load these values into the in-memory cache the next time the user visits our application. We'll add that functionality in the next chapter.
 
 ### Add `Authorization` headers to all requests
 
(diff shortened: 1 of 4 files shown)


docs/source/tutorial/mutations.mdx
```diff
@@ -6,137 +6,132 @@ description: Learn how to update data with the useMutation hook
...
-
-```tsx{2-3,5-10,13-14}:title=src/pages/login.tsx
-export default function Login() {
-  const client: ApolloClient<any> = useApolloClient();
-  const [login, { loading, error }] = useMutation<LoginTypes.login, LoginTypes.loginVariables>(
-    LOGIN_USER,
-    {
-      onCompleted({ login }) {
-        localStorage.setItem('token', login as string);
-        client.writeData({ data: { isLoggedIn: true } });
```
benjamn, line 78: I'm glad to see `client.writeData` go away, but do we think there might be value in preserving the `Query.isLoggedIn` field, perhaps using a custom `read` function? Or is that too far beyond the scope of this material?

docs/source/tutorial/mutations.mdx
```diff
@@ -6,137 +6,132 @@ description: Learn how to update data with the useMutation hook
...
+```tsx:title=src/pages/login.tsx
+const [login, { loading, error }] = useMutation<
+  LoginTypes.Login,
+  LoginTypes.LoginVariables
+>(
+  LOGIN_USER,
+  {
+    onCompleted({ login }) {
+      localStorage.setItem('token', login.token as string);
+      localStorage.setItem('userId', login.id as string);
```
benjamn, line 105:
```suggestion
    onCompleted({ login: { token, id }}) {
      localStorage.setItem('token', token as string);
      localStorage.setItem('userId', id as string);
```
I'm not necessarily proposing this specific change, but I'm curious if you have any strong feelings about these different styles?

benjamn (approved):
LGTM, where G stands for "Great"—thanks @StephenBarlow!

docs/source/tutorial/mutations.mdx
StephenBarlow: Great question! In the "final" version of the tutorial source, this call has been replaced with using a reactive variable, which I figure works better as being added to the _final_ chapter. I don't mind moving that info because the tutorial already requires some steps in the final chapter to get this login flow working.

docs/source/tutorial/mutations.mdx
StephenBarlow: No strong feelings, just matching what's in the tutorial source. Happy to tweak in both spots if preferable.

PR #1035, opened 2020-10-26
Tutorial 4 Basics: Incorrect login query example

quizzy:
#### Issue
Apollo Basics Tutorial part 4 - [Write Mutation Resolvers](https://www.apollographql.com/docs/tutorial/mutation-resolvers/), gives an example for running the login query to [obtain a token](https://www.apollographql.com/docs/tutorial/mutation-resolvers/#obtain-a-login-token):
```gql
mutation LoginUser {
  login(email: "daisy@apollographql.com") {
    email
    token
  }
}
```

When run in the apollo gql playground an error is returned: 
```json
"message": "Field \"login\" must not have a selection since type \"String\" has no subfields.",
``` 
The error occurs because up to this point, the tutorial has set the `login` schema with only a return value of a `String`
```gql
type Mutation {
      bookTrips(launchIds: [ID]!): TripUpdateResponse!
      cancelTrip(launchId: ID!): TripUpdateResponse!
      login(email: String): String
}
```

#### Fix
Update the example to only run a mutation requesting a `String` returned: `login(email: "dasiy@apollo.com")` and the subsequent result that you will see in the apollo-gql playground.


StephenBarlow:
Thank you @quizzy! Looks like we introduced some inconsistencies between the `final` code and the tutorial steps, which I'll resolve separately.

quizzy:
No Probs @StephenBarlow - Thanks for the tutorial 👍

PR #1044, opened 2020-11-06
Add trace information to the Explorer article

daniman:


studio-docs/source/explorer.mdx
```diff
@@ -193,6 +193,23 @@ You can copy responses from your operations with a button or download any given
 
 ### Testing operations
 
+#### Tracing
+
+> Enable **Inlined traces** from the Explorer's Settings tab.
+
+If you are using Apollo Server 2.18+, you can see traces from your responses inlined in the Explorer by adding the `ApolloServerPluginInlineTrace` plugin to your server's configuration.
+
+Turning on the **Inlined traces** feature in the Explorer's settings will tell the Explorer to forward a special header to your server to request that it returns a protobuf with tracing information along with its response.
```
StephenBarlow, line 202: "the Inlined traces feature in the Explorer's settings" -> "inlined traces" (no bold necessary, and the location of the setting is described at the top)

"will tell" -> "tells"

"returns a protobuf with tracing information" -> "return tracing information" (ideally we avoid mentioning protobuf outside of the tracing protocol's definition)

studio-docs/source/explorer.mdx
```diff
@@ -193,6 +193,23 @@ You can copy responses from your operations with a button or download any given
...
+
+```
```
StephenBarlow, line 204: Move this code block between the first para and the second, and end the first para in a colon

PR #1048, opened 2020-11-07
Fix login mutation response type

GhassenRjab:


StephenBarlow:
Thanks for this one as well @GhassenRjab! Was on my to-do list for Monday 😄

PR #1052, opened 2020-11-11
Add dev graph support to full-stack tutorial

StephenBarlow:


docs/source/tutorial/production.md
```diff
@@ -14,6 +14,8 @@ Studio's core features are free for everyone. All of the features in this tutori
 
 ## Create an Apollo account
 
+> Skip this if you already created an account to use the Apollo Studio Explorer.
+
 Visit [studio.apollographql.com](https://studio.apollographql.com) and click **Create an account**. You can sign up either with your GitHub account or by setting a username and password.
```
daniman, line 19: I wonder if we should point everyone to https://studio.apollographql.com/dev for signup here? That route has a streamlined sign up form, compared to the regular one which will have you make an account, invite people, and choose if you want to start a trial.

docs/source/tutorial/resolvers.md
```diff
@@ -99,11 +99,12 @@ By providing your resolver map to Apollo Server like so, it knows how to call re
 
 ## Run test queries
 
-Let's run a test query on our server! Return to GraphQL Playground by running `npm start` and visiting `http://localhost:4000/` in your browser:
+Let's run a test query on our server! Start it up with `npm start` and open whichever tool you previously used to [explore your schema](./schema/#explore-your-schema):
 
-<img class="screenshot" src="../images/playground.png" alt="GraphQL Playground"></img>
+* The Explorer in [Apollo Studio](https://studio.apollographql.com/)
```
daniman, line 104: You'll want to add a `/dev` here I think. It has special behavior:
- logged out, brand new user will be taken to the streamlined signup form
- logged in user with no dev graphs will be taken to the new graph modal
- logged in user with at least one dev graph will be taken to the last dev graph they were looking at

docs/source/tutorial/schema.md
```diff
@@ -204,11 +204,41 @@ server.listen().then(({ url }) => {
...
-> Note that introspection should be **disabled** for a production GraphQL server. Apollo Server disables introspection automatically if the `NODE_ENV` environment variable is set to `production`.
+> Introspection is a helpful GraphQL feature that enables you to obtain a server's schema for development purposes. It should be **disabled** for servers running in production. Apollo Server disables introspection automatically if the `NODE_ENV` environment variable is set to `production`.
+
+### With Apollo Studio (recommended)
+
+The **Apollo Studio Explorer** is a powerful free web IDE for exploring your GraphQL schema and building queries against it:
+
+<img class="screenshot" src="../img/explorer-tab.jpg" />
+
+To use the Explorer with your locally running server, you register a **dev graph** with Apollo Studio, like so:
```
daniman, line 219: Do we want to use the word "register" here? Maybe saying "create" will feel a bit lighter? We've generally been avoiding trying to introduce the registry concept with dev graphs to keep things as simple as possible.

docs/source/tutorial/schema.md
```diff
@@ -204,11 +204,41 @@ server.listen().then(({ url }) => {
...
+
+<img class="screenshot" src="../img/explorer-documentation-tab.jpg" alt="Explorer documentation tab" width="200"/>
+
+Your schema's root types (`Query` and `Mutation`) appear. You can click any type to view its fields and step down into your schema. You'll see that the types and fields match what we provided to our server.
+
+> [Learn more about the Apollo Studio Explorer](https://www.apollographql.com/docs/studio/explorer/)
+
+### With GraphQL Playground
+
+With your server running, visit `localhost:4000` in your browser to open [GraphQL Playground](https://www.apollographql.com/docs/apollo-server/features/graphql-playground/).
```
daniman, line 241: I wonder if we might want to say something about this being a good option for simple needs, but lacking a substantial number of the features/tools in the Explorer?

daniman (approved):
Thanks so much @StephenBarlow ! Had a few questions, but happy to ship this whenever you're ready :)

Now that dev graphs are out of `beta`, we should probably go back and update our "early access" widget that we'd added to the Getting Started tutorial to talk about them as a full entity.

I was also thinking it might be helpful if we added some documentation about what dev graphs are explicitly to Studio's "What is a Graph" doc?

docs/source/tutorial/production.md
StephenBarlow: I'm going to leave this as-is for this PR because it impacts directions for the remainder of the chapter, but we can revisit afterward

PR #1055, opened 2020-11-15
Update production.md tutorial

ivasilyev1:
Working through the tutorial for the first time and was thrown off by the API Key portion because the option is not present unless the Graph Type is "Deployed". Everything up to this point in the tutorial had been using the Development graph type so I was incorrectly trying to do the same for this.

I think a clearer indication that the intent is to use the Deployed Graph Type for this portion of the tutorial would be helpful.


StephenBarlow:
Thanks @ivasilyev1! Development graphs are a cool new thing that we forgot to account for here. I've pushed a fix with #1064.

PR #1062, opened 2020-11-19
Update production.md

brandonleboeuf:
This instruction sequence was missing instructions on which Graph Type to select. This was added as step 3.


StephenBarlow:
Thanks @brandonleboeuf! Development graphs are a cool new thing that we forgot to account for here. I've pushed a fix with #1064.

PR #1071, opened 2020-11-28
Fix capitalization of auto-generated types

catherineluse:
Prevents a compilation error


StephenBarlow:
Thanks @catherineluse! I've incorporated this with #1079

PR #1076, opened 2020-12-09
fix typo in resolver.md

AkatQuas:
There is no `getLaunchIDsByUser ` in the  `final/server/src/datasources/user.js`.

Just a typo.


StephenBarlow:
Thanks for the catch @AkatQuas!

PR #1077, opened 2020-12-11
Fix url bug in docs

saintmalik:
## Summary

Easy navigation and easy reading for others


StephenBarlow:
Thanks for the catch @saintmalik!

PR #1078, opened 2020-12-11
Minor comment: Change in grammar

sudheerhusky:


StephenBarlow:
Thanks for the PR @sudheerhusky! Both options feel valid here so I will leave as-is.

PR #1091, opened 2021-01-01
Fix url bug in docs

saintmalik:
## Summary

Easy navigation and easy reading for others


StephenBarlow:
Thanks very much for all the link fixes @saintmalik! This one will require an absolute link because it's content that's shared across multiple articles. I'll tackle that in a separate PR.

saintmalik:
Okay, thanks

PR #1092, opened 2021-01-05
Update schema.md added scalar Float for completeness

SarathR:
Added scalar Float for completeness


StephenBarlow:
Thanks @SarathR!

PR #1129, opened 2021-03-24
Update roles documentation for billing managers 

Jephuff:
**this now includes the consumer update and should be merged with the other free consumer seat PRs**
`Billing Manager` are now available to all, this updates the [docs](https://www.apollographql.com/docs/studio/org/members) to not assume team plans have all the same role, and not say to contact support to get billing managers

before|after
-|-
image|image


studio-docs/source/org/members.mdx
```diff
@@ -7,13 +7,9 @@ A Studio organization can have any number of members, and each member can be ass
 
 ## Organization-wide member roles
 
-> **Member roles are available for organizations with an Enterprise plan.**
->
-> Apollo Support can also grant complimentary access to the Billing Manager role for Team accounts on request. Please contact **support@apollographql.com** for more information.
+> **Billing Mananger and Org Admins are available to all organizations, the other roles are available for organizations with an Enterprise plan.**
```
daniman, line 10: typo `Mananger` –– maybe you meant `Banananger`? jk :P

We might want to take the other approach here and say our `Consumer` and `Org Admin` roles are available to all accounts, but the more granular roles of `Observer`, `Contributor` and `Graph Admin` are only available ot organizations with an Enterprise plan.

I don't think we need to call out `Billing Manager` as explicitly, that one will be reasonably assumed to be free.

daniman (approved):

studio-docs/source/org/members.mdx
Jephuff: oops! Banananger :D

I merged the other PR into this one, sooo they can go out together and updated the text to closer to what you said! 👍

PR #1136, opened 2021-04-05
Clarify relationship between dataSources and context

SamLee514:
I was a bit confused when I got to the part in chapter 4 (Write query resolvers) that defines the context object without mention of dataSources, even though we expected dataSources to be in the context argument in the query resolvers. Had to check the apollo-server documentation to clear it up. This just adds a quick clarification point to that end. Thanks!

### DIFF
--- a/docs/source/tutorial/mutation-resolvers.md
+++ b/docs/source/tutorial/mutation-resolvers.md
@@ -58,7 +58,9 @@ const server = new ApolloServer({
 });
 ```
 
-The `context` function defined above is called once for _every GraphQL operation_ that clients send to our server. The return value of this function becomes the [`context` argument](./resolvers/#the-resolver-function-signature) that's passed to every resolver that runs as part of that operation (you might have noticed that `dataSources` is nowhere to be found, even though we expect it to be part of the `context` argument in our query resolvers. This is because although we defined `dataSources` outside of `context`, it is [automatically included](https://www.apollographql.com/docs/apollo-server/api/apollo-server/#dataSources) for each operation).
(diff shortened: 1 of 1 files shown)


docs/source/tutorial/mutation-resolvers.md
```diff
@@ -58,7 +58,7 @@ const server = new ApolloServer({
 });
 ```
 
-The `context` function defined above is called once for _every GraphQL operation_ that clients send to our server. The return value of this function becomes the [`context` argument](./resolvers/#the-resolver-function-signature) that's passed to every resolver that runs as part of that operation.
+The `context` function defined above is called once for _every GraphQL operation_ that clients send to our server. The return value of this function becomes the [`context` argument](./resolvers/#the-resolver-function-signature) that's passed to every resolver that runs as part of that operation (you might have noticed that `dataSources` is nowhere to be found, even though we expect it to be part of the `context` argument in our query resolvers. This is because although we defined `dataSources` outside of `context`, it is [automatically included](https://www.apollographql.com/docs/apollo-server/api/apollo-server/#dataSources) for each operation).
```
StephenBarlow, line 61: Great clarification! Could you reformat this to be a separate blockquote below the paragraph instead:

```
> You might have noticed that...
```

StephenBarlow:
Thanks for the PR @SamLee514! Had a small suggestion for formatting. Will be happy to merge after that's incorporated 🎉

SamLee514:
@StephenBarlow reformatted!

StephenBarlow:
Thanks @SamLee514, looks great! Merged.

PR #1138, opened 2021-04-07
Build webhooks preview docs

jsegaran:


studio-docs/source/composition-publish-notification.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Composition Publish Webhook (Preview)
```
jstjoe, line 2: 1. adjust the title capitalization to match the other titles (lower case after the first term)
2. Let's use the same name as in the payload: Composition status

So "Composition status notifications (preview)".

jstjoe (approved):
LGTM. We could add some concrete examples to make it even better but this is good to go.

PR #1142, opened 2021-04-16
Fix: book-trips.tsx error

Barrosg99:
Missing '!' causes "Variable $launchIds of type [ID]! used in position expecting type [ID!]." error on book-trips.tsx


StephenBarlow:
Hi @Barrosg99, I believe you might have put a `!` in the incorrect spot in your `bookTrips` schema definition? This is the definition as written in the "Build a schema" chapter:

```
type Mutation {
  bookTrips(launchIds: [ID]!): TripUpdateResponse!
  cancelTrip(launchId: ID!): TripUpdateResponse!
  login(email: String): User
}
```

In terms of what the tutorial is expecting by default, I believe this is correct as-is 🤔

PR #1143, opened 2021-04-19
Initial work on composition check doc updates

StephenBarlow:


studio-docs/source/schema-checks.mdx
```diff
@@ -88,23 +122,48 @@ You override flagged changes on an operation-by-operation basis. For each operat
...
+If you have a [federated graph](https://www.apollographql.com/docs/federation/), you run schema checks with the `rover subgraph check` command (instead of `rover graph check`):
 
-If your data graph uses [Apollo Federation](https://www.apollographql.com/docs/federation/), the schema checks process differs slightly. For details, see [Checking changes to a federated graph](https://www.apollographql.com/docs/federation/managed-federation/advanced-topics/#checking-changes-to-a-federated-graph).
+```bash
+rover subgraph check docs-example-graph@current --schema ./schema.graphql --name products
+```
+
+In addition to the arguments required by `graph check` , `subgraph check` requires the `--name` of the particular subgraph you're checking against.
+
+When you run `subgraph check`, Apollo Studio performs operation checks as usual, along with **composition checks**: Studio verifies that the changes you make to a subgraph schema are compatible with your _other_ subgraph schemas, enabling them to compose into a _supergraph_ schema ([learn more](https://www.apollographql.com/docs/federation/quickstart/#federation-concepts)).
```
caydie-tran, line 147: This is looking awesome! The one thing to call out here is that for a federated graph, we actually run composition checks _before_ operation checks. So if composition fails, then we just don't run operation checks. We should add this bit of information somewhere here :)

studio-docs/source/schema-checks.mdx
```diff
@@ -88,23 +122,48 @@ You override flagged changes on an operation-by-operation basis. For each operat
...
+
+
+Results of both check types are available in Studio from your graph's Checks tab:
+
+<img class="screenshot" src="./img/schema-checks/checks-studio.jpg" alt="Schema check results in Studio" width="300" />
+
+You can click on a composition check to view its result. If composition succeeded, you can view the composed supergraph schema.
```
caydie-tran, line 154: We can also mention that if composition check fails, then you can view the subgraph schema

PR #1155, opened 2021-04-30
chore: adds link to rover docset

EverlastingBugstopper:
i also alphabetized these links since I wasn't quite sure where to put rover in the list - lemme know if there was a specific ordering we wanted to keep.


StephenBarlow (approved):
Alphabetical works for me here!

PR #1158, opened 2021-04-30
First set of Studio docs changes for Rover GA

StephenBarlow:


studio-docs/source/github-integration.md
```diff
@@ -23,14 +23,14 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:8
+      - image: cimg/node:16.0.0
```
lrlna, line 26: yep that would be a useful update to have :joy:

lrlna (approved):
LGTM!

studio-docs/source/github-integration.md
```diff
@@ -23,14 +23,14 @@ version: 2
...
 
     steps:
       - checkout
 
       - run: npm install
       # CircleCI needs global installs to be sudo
-      - run: sudo npm install --global apollo
+      - run: sudo npm install --global @apollo/rover
```
abernix, line 33: I think we were previously fighting peer dependencies and overlapping `graphql` versions, but since we're done with that with the introduction of Rover, I think in a future / soon change to these instructions it would be advisable for users to directly install `@apollo/rover` in their application's `devDependencies`, and then merely call `npx rover` here from their project (after the `npm install`).

That would provide them a pinned version and reproducible build dependencies by avoiding always just installing the _`latest`_ version of `@apollo/rover`.  (Also, presumably, their developers would appreciate having `npx rover` available to them.  

I wonder where would be a good place to track this work?  Would this leverage an installation instruction best-practice elsewhere?  Or would it just make the suggestion here itself?

studio-docs/source/org/members.mdx
```diff
@@ -57,7 +57,7 @@ When `Contributors` in an organization create a new graph, they are automaticall
 | Manage Explorer settings (URL, etc)                             |     ✓     |      ✓      |  Non-protected variants only      |          |          |
 | Create [Deployed Graphs](/org/graphs/)                          |     ✓     |      ✓      |      ✓      |          |          |
 | Run schema checks                                               |     ✓     |      ✓      |      ✓      |    ✓     |          |
-| See the schemas of implementing services in federated graphs    |     ✓     |      ✓      |      ✓      |    ✓     |          |
+| See the schemas of subgraphs in federated graphs    |     ✓     |      ✓      |      ✓      |    ✓     |          |
```
abernix, line 60: Did we want to keep these markdown tables aligned? (I could understand either way, it just looks aligned right now.)

```suggestion
| See the schemas of subgraphs in federated graphs                |     ✓     |      ✓      |      ✓      |    ✓     |          |
```

abernix (approved):
LGTM! I have one _deeper_ thought about our recommended pattern here resulting in non-reproducible builds since the version of `@apollo/rover` isn't _pinned_ in any way, but that was the case before.  (I just think that _now_ we can avoid the reason we may have danced around that suggestion before, which I noted in the comment.)

Thanks for this!

studio-docs/source/github-integration.md
StephenBarlow: Good idea! I'll make "general CI tidying" a near-future focus for nuances like this.

PR #1188, opened 2021-06-19
Changes to request headers in Explorer

esilverm:
We no longer support using JSON for setting or storing request headers.


StephenBarlow (approved):
Sorry I didn't see this sooner @esilverm!

PR #1190, opened 2021-06-28
Update data-source.md blog link

corinagum:
The old URL led to a 404, so I googled the title and this is the link that I'm pretty sure replaces it.


StephenBarlow:
Thanks @corinagum!

PR #1195, opened 2021-07-03
(docs) @contact directive for schemas with a Mutation type

setchy:
Required to prevent @apollo/rover errors when publishing to the schema registry /w managed federation

PR #1196, opened 2021-07-06
mention other root operation types in contact directive definition

StephenBarlow:


StephenBarlow:
Merged as alternative to https://github.com/apollographql/apollo/pull/1195

PR #1222, opened 2021-08-26
Embeddable Explorer in the Apollo Docs

mayakoneval:
https://user-images.githubusercontent.com/14367451/131048072-db209424-ba0d-4405-aa7f-b28d9ce48199.mov

**This PR**
Setup an embedded explorer component in the docs fetching from Acephei. 

I didn't try very hard on this copy, just wanted to make the `EmbeddedExplorer` component in the docs, since I assume the people who are good with words would be focusing on the copy. 

For blog posts, I assume we can reuse components from the docs, so I think the bulk of the work there is with the word smiths like y'all. :appreciation:

**please feel free to commit directly to this branch! I just wanted to set up the embeddable component to work**

**Questions to consider**
- How should we show that the EMBEDDED_EXPLORER_URL should be changed from user to user depending on your url?
- Is the code snippet too long? Should it be an [expandable panel](https://www.apollographql.com/docs/studio/getting-started/#3-register-your-schema) instead?
- Should we pop out an embedded explorer into a modal so folks can see it better? Or should we have a dedicated page for the embedded explorer?

**TODO**
- ~switch the embedded explorer url to the prod version once the public variants pr is merged~
- blog post embed
- ~subscriptions example (requires opening a connection and closing it when a termination post message is received) (also acephei doesn't support subscriptions right now, so not sure if we want to add that to acephei or use the other subscription app)~


mayakoneval:
> This is so dang exciting!
> 
> Instead of defining the component here, let's define it in the `gatsby-theme-apollo-docs` package, in this directory: https://github.com/apollographql/gatsby-theme-apollo/tree/main/packages/gatsby-theme-apollo-docs/src/components
> 
> Let's also generic-ify it to take an embed URL as a prop so we can use it all over the place talking to different graphs 🎉
> 
> Once we get the theme merged and updated in our docsets, we can import it here in the Studio docs w/ `import {EmbeddedExplorer} from 'gatsby-theme-apollo-docs';`

@StephenBarlow here is that component: https://github.com/apollographql/gatsby-theme-apollo/pull/220

I also added subscriptions support with `subscription-transport-ws`

trevorblades:
Awesome! I just merged and published your embeddable explorer PR. You should now be able to do:

```js
import {EmbeddableExplorer} from 'gatsby-theme-apollo-docs';

<EmbeddableExplorer
  graphRef="123"
  graphEndpoint="abc"
  graphSubscriptionEndpoint="def"
  styles={{width: '100%'}}
/>
```

PR #1246, opened 2021-10-07
chore: don't recommend installing Rover w/sudo

EverlastingBugstopper:
Rover installers should never be run with `sudo`


studio-docs/source/schema/cli-registration.mdx
```diff
@@ -37,8 +37,15 @@ jobs:
...
 
       - run: npm install
-      # CircleCI needs global installs to be sudo
-      - run: sudo npm install --global @apollo/rover
+      
+      - run:
+          name: Install Rover
+          command: |
+            # download and install Rover
+            curl -sSL https://rover.apollo.dev/nix/v0.3.0 | sh
```
StephenBarlow, line 45: @EverlastingBugstopper Oops sorry I totally let this PR lie fallow 😞 

Is this version string going to be changing a lot? We might at least want to expand the comment to indicate to check what version you're using locally or whatever makes sense

PR #1257, opened 2021-10-21
How contract supergraph schemas work

zionts:
Clarify that contract gateways can still use elided fields for internal routing and that @tag directive labels are maintained

### DIFF
--- a/studio-docs/source/contracts.mdx
+++ b/studio-docs/source/contracts.mdx
@@ -44,7 +44,7 @@ graph BT;
 
 This enables you to hide experimental types and fields that are still in development, or to limit a particular audience's access to only the portions of your graph that they need.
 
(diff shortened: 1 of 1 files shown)


studio-docs/source/contracts.mdx
```diff
@@ -44,7 +44,7 @@ graph BT;
 
 This enables you to hide experimental types and fields that are still in development, or to limit a particular audience's access to only the portions of your graph that they need.
 
-Contract gateways can safely connect to the same subgraph instances as any other gateway, because they can only interact with data that's represented in the contract schema.
+Contract gateways can safely connect to the same subgraph instances as any other gateway, because they can only interact with data that's represented in the contract schema. This will not affect internal routing (e.g. usage of hidden fields in `@requires` selection sets). Additionally, any `@tag` labels that are part of the source variant's supergraph schema will be maintained on a contract supergraph schema.
```
StephenBarlow, line 47: "will not" -> "does not"

(e.g. ...) -> (e.g., using excluded fields in a `@requires` directive)

"will be maintained on a" -> "are preserved in a"

PR #1285, opened 2021-12-18
contracts: Fix `@tag` definition

david-castaneda:
Fix the @tag definition in the contracts example by removing the `|` before `FIELD_DEFINITION`

I used this formatter but not sure if theres a recommended way to format it:
https://jsonformatter.org/graphql-formatter

### DIFF
--- a/studio-docs/source/contracts.mdx
+++ b/studio-docs/source/contracts.mdx
@@ -81,11 +81,9 @@ For example, let's take a look at this Products subgraph schema:
 
 ```graphql{1-6,18-21}:title=products.graphql
 # You must include this definition in any schema with tags!
-directive @tag(name: String!) repeatable on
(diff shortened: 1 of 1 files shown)


david-castaneda:
Interestingly, javascript parses both definitions correctly, the one with the `|` before and after `FIELD_DEFINITIONS`

And also, the spec @tag definition matches the current documentation
https://specs.apollo.dev/tag/v0.1/#sel-FAFFBADAAACeB7zE

However, given that the @tag definition in [federation core tests]( https://github.com/apollographql/federation/blob/5c236c5cb50f911de2d57c5a9d954662724f281c/gateway-js/src/core/__tests__/core.test.ts#L32-L34) and internal studio logic has the `|` before the type removed then i've updated the docs to reflect that.

PR #1303, opened 2022-01-26
Update preflight script docs with API

Jephuff:
Add documentation for the different helpers in preflight scripts
image

### DIFF
--- a/studio-docs/components/InlineCode.js
+++ b/studio-docs/components/InlineCode.js
@@ -1,0 +1,5 @@
+export const InlineCode = ({ children }) => (
+    <code className="language-text">
+        {children}
+    </code>
+);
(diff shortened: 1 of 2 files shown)


papollosc:
Tis dope. I would rely on @StephenBarlow for language clarity. Wondering if we want to add the snippet titles (`Send request`, `get variable`, `set variable`, `Request user input`)  so it is easy to recognize? But that would make it four columns, not sure if that is encouraged. 
image

papollosc (commented):
Tis dope. I would rely on @StephenBarlow for language clarity. Wondering if we want to add the snippet titles (`Send request`, `Get variable`, `Set variable`, `Request user input`)  so it is easy to recognize? But that would make it four columns, not sure if that is encouraged. 
image

StephenBarlow:
Oh hey nice! We have a special table format that we use for API references such as these for extra snazziness. Tomorrow I'll put together a PR against _this_ PR to convert it, plus any language tweaks.

papollosc (approved):
🥳 🙇‍♀️

PR #1305, opened 2022-01-28
Edits to preflight edits

StephenBarlow:


StephenBarlow:
@Jephuff I think we can get away with this snazzy table format _without_ a new docset-specific component, if that works for you! Let me know if any of my wording changes or anything could use work 😄

Jephuff:
> @Jephuff I think we can get away with this snazzy table format _without_ a new docset-specific component, if that works for you! Let me know if any of my wording changes or anything could use work 😄

haha using an actual table didn't even cross my mind to get around the escaping | hahaha this works! thanks! :)

Jephuff (approved):

PR #1320, opened 2022-03-03
Update references to op reg plugin

trevor-scheer:
The `apollo-server-plugin-operation-registry` package is now being published under the name `@apollo/server-plugin-operation-registry`.

The related docs / migration guide that existed for the `0.2.x` versions could probably use a refresher (or maybe removal, though a quick perusal of npm downloads per version shows that some people are still using _very_ old versions of the plugin and would qualify for the migration guide).

cc @StephenBarlow @rkoron007


glasser (approved):

rkoron007 (approved):
Looks good to me! :)

